PR #6, opened 2021-10-14
add Pivot transform with dynamic 'runquery' for distinct vals

cpdough:


Transforms/table_operations/rasgo_pivot.sql
```diff
@@ -1,3 +1,13 @@
...
+{%- set get_distinct_vals -%}
+    select distinct {{ value_column }}
+    from {{ source_table }}
+    limit 1000
+{%- endset -%}
+{%- set results = run_query(get_distinct_vals) -%}
+{%- set distinct_vals = results[results.columns[0]].to_list() -%}
+
+SELECT {{ dimensions }}, {{ distinct_vals | join(", ") }}
+FROM ( SELECT {{ dimensions }}, {{ pivot_column }}, {{ value_column }} FROM {{ source_table }})
```
bstempi, line 12: We probably want to get into the habit of quoting our column names to avoid issues down the road. Something like:

```
"{{ distinct_vals | join('", "') }}"
```

cpdough:
the new commit where i tried to make the value lookup optional above doesn't work (error message below). would love some help if you have any ideas @alexwolf22 @bstempi 

`Internal Error Details: {'message': 'Error parsing template to query string', 'error': 'cannot unpack non-iterable NoneType object'}`

Transforms/table_operations/rasgo_pivot.sql
bstempi: Having a special jinja filter would be super cool and probably make writing this stuff much easier.

bstempi:
I think you're looking for:

`{%- if list_of_vals is defined -%}`

PR #11, opened 2021-10-21
progress on auto docs generation

alexwolf22:
- New module `docs/` containing automatically generated markdown docs based on transform YAMLs
   -  This will then be copied in pasted it out [Rasgo Docs Repo](https://github.com/rasgointelligence/RasgoDocs) ( Or hopefully make it manual in the future)
- Python package at `docs/docs_generation/` for generating the docs automatically
    - Note need to install packages in `docs/docs_generation/requirements.txt` 
  
## Running the Docs Generator
1. Create a **venv** with the requirements.txt file
2. Activate it
3. Call the python file like below
```
STILL FIGURING OUT THE BEST WAY
```

## TODO
- Make git hook to run the process on every commit

## Important Notes
- We are separating the markdown files in direct transform type dirs  in the `docs` folder (won't all just be in that file)
- Process reads every YAML file and overwrites the markdown file, regardless of if there is one there or not


docs/docs_generation/doc_generator.py
```diff
@@ -0,0 +1,59 @@
...
+
+def _get_transform_markdown(transform_data: Dict,
+                            transform_type: str,
+                            transform_name: str) -> str:
+    """
+    Generate and return the markdown string to write as a MD
+    file for this transform based off the YAML file
+    """
+    # Generate Markdown Elements in Transform Doc
+    markdown_elements = [
```
bstempi, line 43: I'm surprised you didn't use Jinja here 😄

PR #12, opened 2021-10-22
moving average

cpdough:
adding moving_average UDT. thanks @engelAnalytics


cpdough:
@alexwolf22  can you double-check this one? i think its ready

bstempi (approved):
LGTM. I had a single comment about naming, but that's definitely not something that should stop this from merging.

PR #27, opened 2021-11-04
Create transform args

alexwolf22:
Python script to create all transforms in the UDT repo with PyRasgo.

The command takes as a positional argument, the PyRasgo API key, and if your API KEY isn't a production one not you'll need to supply the Rasgo domain with the `-d` option (local, staging, or production which is the default).

Script works by
- Reading all YAML/Source code files in each of the UDT Transform Dictroies (row, column, and table)
- If transform with that name exists in DB already don't create, else create it

This is just an MVP, and changes can be made later.

Organization of transform doc generation code is also in this PR so all python code lives in the (python)[./python] module.

## Python Env setup
Yeah will need two libraries installed
- `pyrasgo>=0.3.3a3`
- `pyyaml`

You could use the [requirments.txt](./python/requirements.txt) file to install the needed libraries or do it in your own environment.

## Script Execution
Execute it like below
```
python python python/create_transforms.py <pyrasgo-api-key> -d <local, staging, or production (default)>
```


python/create_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+                transform_source_code = utils.get_transform_source_code(
+                    transform_type=transform_type,
+                    transform_name=transform_name
+                )
+                transform_args = utils.parse_transform_args_from_yaml(transform_yaml)
+
+                # Create Transform
+                print(f"Creating '{transform_name}' transform in Rasgo {rasgo_domain.upper()} "
+                      f"environment, with args {transform_args}")
+                rasgo.create.transform(
```
bstempi, line 49: If this call fails, do we want it to immediately throw an exception (which it does now), or would we rather it log the issue and continue on?

python/create_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+                    name=transform_name,
+                    source_code=transform_source_code,
+                    arguments=transform_args
+                )
+
+
+if __name__ == "__main__":
+    arg_parser = argparse.ArgumentParser()
```
bstempi, line 57: Thank you for using arg parser like a proper adult and not trying to parse the args by hand.

python/utils.py
```diff
@@ -45,6 +45,72 @@ def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
...
+    os.environ["RASGO_DOMAIN"] = rasgo_domain_url
+
+
+def get_all_rasgo_transform_names(rasgo: Rasgo) -> Set[str]:
+    """
+    Return a list of all transforms names available to
+    the logged in Rasgo User
+    """
+    transform_in_rasgo = rasgo.get.transforms()
+    return {t.name for t in transform_in_rasgo}
```
bstempi, line 80: 👍  for set notation.

bstempi (approved):
Overall, LGTM. Any comments made were either questions or things that were so small that they shouldn't hold this up.

column_transforms/rasgo_one_hot_encode/rasgo_one_hot_encode.sql
```diff
@@ -0,0 +1,34 @@
+{#
```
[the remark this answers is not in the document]
alexwolf22: No that was an accident removed it

python/constants.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+from enum import Enum
+
+
+class PyRasgoEnvironment(Enum):
+    """
+    Different Environment for connecting PyRasgo to
+    """
+    PRODUCTION = "api.rasgoml.com"
+    STAGING = "staging-rasgo-proxy.herokuapp.com"
```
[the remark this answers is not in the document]
alexwolf22: I just copied the URLs for what we had in PyRasgo. Pretty sure it works.

python/create_transforms.py
alexwolf22: PyRasgo throws the exceptions and raises proper error messages already so don't need to do it here

bstempi: That's not the point I was raising.

"If 1 of these operations fail, do you want to keep trying the rest of them?" was what I was getting at.

PR #28, opened 2021-11-05
added new math transform

cpdough:
what y'all think of this approach for math ops?

also cleaned typos in rasgo_filter @alexwolf22


bstempi (approved):
It's super simple. I dig it.

PR #30, opened 2021-11-05
Find and replace udt

alexwolf22:
Made find and replace UDT. Works by supplying a column/s and a nested list of find + replace vals, and will do the replacements.

For example

```python
source.transform(
    transform_name=transform_name,
    replace_dict={
        "MONTH": [[4, 44], [2, 22]],   # Replace value 4 with 44 in MONTH column. Also replace 2 with 22.
        "FIPS": [['45001', '455001']]  # Replace value '45001' with '455001' in FIPS column.
    }
).preview()
```

will produce

```sql
SELECT
    YEAR,
    COVID_DEATHS,
    DATE,
    COVID_NEW_CASES,
    CASE
        WHEN FIPS = '45001' THEN '455001'
        ELSE FIPS
    END as FIPS,
    CASE
        WHEN MONTH = 4 THEN 44
        WHEN MONTH = 2 THEN 22
        ELSE MONTH
    END as MONTH,
    COVID_CUMULATIVE_CASES
FROM RASGOCOMMUNITY.PUBLIC.COVID_MONTHLY_FEATURES
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+.DS_Store
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/column_transforms/rasgo_find_and_replace/rasgo_find_and_replace.sql
+++ b/column_transforms/rasgo_find_and_replace/rasgo_find_and_replace.sql
@@ -11,10 +11,23 @@ the columns in a table by fqtn
     AND   TABLE_NAME = '{{ table }}'
 {%- endmacro -%}
 
+{# Jinja Macro to single quote a value if a sting #}
 {%- macro quote_if_string(val) -%}
     {{ "'" + val + "'" if val is string else val }}
 {%- endmacro -%}
 
+{#
+Jinja Macro to create a WHEN statement in find and replace considering Nulls
+#}
+{%- macro make_when_statement(col, find_val, replace_val) -%}
+    {%- if find_val -%}
+    WHEN {{ col }} = {{ quote_if_string(find_val) }} THEN {{ quote_if_string(replace_val) }}
+    {%- else -%}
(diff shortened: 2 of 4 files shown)


cpdough:
nice work alex. this is cool but it's not very intuitive. also - it won't support multiple conditions for a replace (like `when col a = 1 and col b is NULL then 5`). i'm wondering if we should use the filter approach where we constructed the args with a list like:
`['COLa =1 AND COLb is NULL', '5']`

cpdough:
we could even add the column in too so its like
`['MONTH', 'COLa =1 AND COLb is NULL', '5']
['MONTH', 'COLa =2', '10']
['FIPS', 'FIPS=100', '10001']
`
then we could collapse the ones that reference the same column into a single case statement

and (last thing) this is probably a good opportunity to keep the original column and name the new column `COLNAME_REPLACED` so that way the user can optionally drop the original column later if they choose

bstempi:
> and (last thing) this is probably a good opportunity to keep the original column and name the new column COLNAME_REPLACED so that way the user can optionally drop the original column later if they choose

Absolutely agreed on this point.

> also - it won't support multiple conditions for a replace (like when col a = 1 and col b is NULL then 5).

I feel like this is something different than a "find and replace." This seems like a more generic column transform.

Do we want this to be something more generic than a "find and replace (within a) column?" I'm not sure what that looks lile.

cpdough:
> I feel like this is something different than a "find and replace." This seems like a more generic column transform.

i'm down with this. we can leave this one as the find and replace and make a separate 'CASE WHEN' transform (or maybe just call it if then else transform) for the more complex, less straightforward scenarios

alexwolf22:
> and (last thing) this is probably a good opportunity to keep the original column and name the new column `COLNAME_REPLACE`

Okay I don't think we have a drop columns UDT yet. Should we make one of these?

PR #32, opened 2021-11-08
add train/test split UDT

cpdough:
shippable like this but should add a RAND option for when theres no date to order by (and make the order by optional so it can be excluded in that scenario)


griffatrasgo (approved):
SQL looks correct. Someone else should review the udt syntax

bstempi (approved):
LGTM

PR #36, opened 2021-11-10
initial changes for some more basic UDTs

szaghikian-rasgo:
This is a draft. I have plenty of work to do - docs and ensuring the actual functionality I want. Questions I could use reviews on:
1. Do we want to offer in-place and new column creation options for the `TRY-CAST` UDT?
2. How do we feel about not creating explicit new column aliases, but instead letting them receive default names from Snowflake based on their creation arguments?

Any other input welcome. Thanks amigos!

### DIFF
--- a/column_transforms/rasgo_cast/rasgo_cast.sql
+++ b/column_transforms/rasgo_cast/rasgo_cast.sql
@@ -1,7 +1,11 @@
+{% if col_list|length != type_list|length %}
+Rasgo Cast Error: The Column list must be the same length as the Type list.
+{% else %}
 SELECT
 *
-{%- for col_expr, new_type in new_cols %}
-    -- TODO: offer an option for inplace or new column for each change?
-    , TRY_CAST({{col_expr}} AS {{new_type}})
+{%- for target_col in col_list %}
+    , TRY_CAST({{target_col}} AS {{type_list[loop.index-1]}})
 {%- endfor %}
 FROM {{ source_table }}
+
+{% endif %}
--- a/column_transforms/rasgo_cast/rasgo_cast.sql
+++ b/column_transforms/rasgo_cast/rasgo_cast.sql
@@ -4,7 +4,7 @@ Rasgo Cast Error: The Column list must be the same length as the Type list.
 SELECT
 *
 {%- for target_col in col_list %}
-    , TRY_CAST({{target_col}} AS {{type_list[loop.index-1]}})
+    , TRY_CAST({{target_col}} AS {{type_list[loop.index-1]}}) AS {{cleanse_name(target_col)+'_'+cleanse_name(type_list[loop.index-1])}}
 {%- endfor %}
 FROM {{ source_table }}
 
--- a/column_transforms/rasgo_cast/rasgo_cast.sql
+++ b/column_transforms/rasgo_cast/rasgo_cast.sql
@@ -1,0 +1,7 @@
+SELECT
+*
+{%- for col_expr, new_type in new_cols %}
+    -- TODO: offer an option for inplace or new column for each change?
+    , TRY_CAST({{col_expr}} AS {{new_type}})
+{%- endfor %}
+FROM {{ source_table }}
--- a/column_transforms/rasgo_cast/rasgo_cast.yaml
+++ b/column_transforms/rasgo_cast/rasgo_cast.yaml
@@ -1,6 +1,6 @@
-name: rasgo_concat
+name: rasgo_cast
 transform_type: column
-source_code: rasgo_concat.sql
+source_code: rasgo_cast.sql
 description: |
   This function creates new columns for all columns included in `col_list`. Each of the new columns is cast to a new data type, indicated in `type_list`.
 
@@ -18,7 +18,7 @@ example_code: |
     rasgo.read.source_data(w_source.id, limit=5)
 
     t1 = w_source.transform(
-        transform_name='rasgo_concat',
+        transform_name='rasgo_cast',
         col_list=["DS_WEATHER_ICON", "DS_DAILY_HIGH_TEMP", "DS_DAILY_LOW_TEMP"],
         type_list=["INT", "STRING", "STRING"]
     )
--- a/column_transforms/rasgo_cast/rasgo_cast.yaml
+++ b/column_transforms/rasgo_cast/rasgo_cast.yaml
@@ -1,0 +1,27 @@
+name: rasgo_concat
+transform_type: column
+source_code: rasgo_concat.sql
+description: |
+  This function creates a new column that contains a substring of either a fixed value or another column in your dataset.
+
+  Caution! Executing this transformation on a large dataset can be computationally expensive!
+  # TODO: finish description
+arguments:
+  new_cols:
+    type: List[string, string]
+    description: A list representing each existing column to be replaced.
+  col_expr:
+    type: string
+    description: The column targeted for data type change
+  new_type:
+    type: string
+    description: The desired output data type for the new column
+example_code: |
+    rasgo.read.source_data(w_source.id, limit=5)
+
+    t1 = w_source.transform(
+        transform_name='rasgo_substring',
+        new_cols = [column_name, SNOWFLAKE_DATA_TYPE]
+      )
+
+    t1.preview()
--- a/column_transforms/rasgo_cast/rasgo_cast.yaml
+++ b/column_transforms/rasgo_cast/rasgo_cast.yaml
@@ -2,26 +2,25 @@ name: rasgo_concat
 transform_type: column
 source_code: rasgo_concat.sql
 description: |
-  This function creates a new column that contains a substring of either a fixed value or another column in your dataset.
+  This function creates new columns for all columns included in `col_list`. Each of the new columns is cast to a new data type, indicated in `type_list`.
+
+  The ordinal position of columns in `col_list` must match the desired output data type in `type_list`.
 
   Caution! Executing this transformation on a large dataset can be computationally expensive!
-  # TODO: finish description
 arguments:
-  new_cols:
-    type: List[string, string]
-    description: A list representing each existing column to be replaced.
-  col_expr:
-    type: string
-    description: The column targeted for data type change
-  new_type:
-    type: string
-    description: The desired output data type for the new column
+  col_list:
+    type: column_list
+    description: A list representing each existing column to have a changed type.
+  type_list:
+    type: value_list
+    description: A list of strings containing the types to which to cast the columns.
 example_code: |
     rasgo.read.source_data(w_source.id, limit=5)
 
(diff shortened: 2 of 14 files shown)


column_transforms/rasgo_cast/rasgo_cast.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+*
+{%- for col_expr, new_type in new_cols %}
+    -- TODO: offer an option for inplace or new column for each change?
```
cpdough, line 4: new columns is fine for now - we've followed that pattern throughout. i'll add a drop columns UDT to the list so we can drop unwanted ones explicitly. eventually we probably make more of them work in place but for now, not a big deal.

column_transforms/rasgo_cast/rasgo_cast.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+*
+{%- for col_expr, new_type in new_cols %}
```
cpdough, line 3: i like this idea but the nested arg types this requires isn't something we have supported in the UI yet... life may be easier here if you just create two lists and force them to be ordinally aligned so like:
column_list: ['col1', 'col2', 'col3']
type_cast: ['string', 'int', 'float']

the other option is to go with a more complex dict but that feels like overkill here. can discuss more live

column_transforms/rasgo_cast/rasgo_cast.yaml
```diff
@@ -0,0 +1,27 @@
+name: rasgo_concat
```
cpdough, line 1: yaml doesn't match the UDT - i think you mentioned this, just flagging in case

column_transforms/rasgo_concat/rasgo_concat.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+*
+{%- for col_expr, vals in new_cols %}
```
cpdough, line 3: i think this one can be simplified by taking one `column_list` as the arg and concatenating them in order that they are supplied

so ['col1', 'col2', 'col3'] would generate a new column col1 || col2 || col3 as ConcatenatedCol

column_transforms/rasgo_substring/rasgo_substring.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+*
+{%- for col_expr, start_pos, end_pos in new_cols %}
```
cpdough, line 3: do we anticipate needing to run multiple substrings on multiple different columns in one UDT?

cpdough (changes requested):

column_transforms/rasgo_cast/rasgo_cast.yaml
```diff
@@ -0,0 +1,27 @@
...
+arguments:
+  new_cols:
+    type: List[string, string]
+    description: A list representing each existing column to be replaced.
+  col_expr:
+    type: string
+    description: The column targeted for data type change
+  new_type:
+    type: string
+    description: The desired output data type for the new column
```
alexwolf22, line 18: In the example code I'm only seeing the param `new_cols` inputted. What about the other two? (`new_type and `col_expr`

column_transforms/rasgo_cast/rasgo_cast.yaml
```diff
@@ -0,0 +1,27 @@
+name: rasgo_concat
+transform_type: column
+source_code: rasgo_concat.sql
+description: |
+  This function creates a new column that contains a substring of either a fixed value or another column in your dataset.
+
+  Caution! Executing this transformation on a large dataset can be computationally expensive!
+  # TODO: finish description
```
alexwolf22, line 8: Remember to remove this once ready to merge

column_transforms/rasgo_cast/rasgo_cast.yaml
```diff
@@ -0,0 +1,27 @@
...
+    type: List[string, string]
+    description: A list representing each existing column to be replaced.
+  col_expr:
+    type: string
+    description: The column targeted for data type change
+  new_type:
+    type: string
+    description: The desired output data type for the new column
+example_code: |
+    rasgo.read.source_data(w_source.id, limit=5)
```
alexwolf22, line 20: I don't think we need this line here. 

Just have `source.....` like we do in other yaml files

column_transforms/rasgo_concat/rasgo_concat.yaml
```diff
@@ -0,0 +1,26 @@
+name: rasgo_concat
```
alexwolf22, line 1: Same comments about YAML file above

column_transforms/rasgo_substring/rasgo_substring.yaml
```diff
@@ -0,0 +1,26 @@
...
+    description: The number of characters to select. If left null, automatically select through the end of the string.
+example_code: |
+    rasgo.read.source_data(w_source.id, limit=5)
+
+    t1 = w_source.transform(
+        transform_name='rasgo_substring',
+        new_cols = [column_name, 1, 5]
+      )
+
+    t1.preview()
```
alexwolf22, line 26: Not seeing all params used in the example

column_transforms/rasgo_substring/rasgo_substring.yaml
```diff
@@ -0,0 +1,26 @@
+name: rasgo_substring
+transform_type: column
+source_code: rasgo_substring.sql
+description: |
+  This function creates a new column that contains a substring of either a fixed value or another column in your dataset.
+
+  # TODO: finish description
+arguments:
+  col_expr:
+    type: string
```
alexwolf22, line 10: I think this should be of type `column` right? We use that type so in the UI we prepopulate a select drop down with all columns in that dataset.

Easier than manually typing in.

alexwolf22 (changes requested):
Make sure we tidy up all the YAML files/example code and descriptions.

In regards to the transform args, they need to be the ones inputted into the transform function. I noticed for some UDTs you had 3 args, but only one inputted into the column (`new_cols`). If you only meant to have one parameter to the transform function, it should only be one in the YAML file describing what each of the three values should be.

If we make a new arg type that's not in this [spreadsheet](https://docs.google.com/spreadsheets/d/1kXC7dn8mcW9cSZUYEJu-1Hl73urad_uL8Lz1BDHia9U/edit#gid=249017404) already, make sure to add it we an example.

We should try to minimize the amount of new transform arg types we create, and use existing ones if possible.

column_transforms/rasgo_cast/rasgo_cast.yaml
szaghikian-rasgo: yeah I didn't really know what to do with the docs. I think they should be more appropriate now. Mind taking another look?

column_transforms/rasgo_cast/rasgo_cast.yaml
szaghikian-rasgo: let's make a call on this? Some files it's just the one call to `source.transform`, some it's the full workflow I've got here. PL tomorrow morning?

table_transforms/rasgo_dedupe/rasgo_dedupe.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+*
+FROM {{ source_table }}
+QUALIFY ROW_NUMBER() OVER (
+    PARTITION BY {%- for col in natural_key %} {{col}}{{"," if not loop.last else ""}} {%- endfor %}
+    ORDER BY {%- for col in order_col %} {{col}}{{"," if not loop.last else ""}} {%- endfor %} {{order_method}}
+) = 1
```
cpdough, line 7: we should specifically say in the docs that once the order by is applied, only the top row will be kept and the rest will be dropped

column_transforms/rasgo_concat/rasgo_concat.yaml
```diff
@@ -0,0 +1,21 @@
...
+arguments:
+  concat_list:
+    type: mixed_list
+    description: A list representing each new column to be created.
+example_code: |
+    rasgo.read.source_data(w_source.id, limit=5)
+
+    t1 = w_source.transform(
+      transform_name='rasgo_concat',
+      concat_list=["DS_WEATHER_ICON", "'some_str'", "'_5'"]
```
cpdough, line 18: will the single quotes be required on non-columns? is there a way to add those in the template so that we don't put that burden on the user?

cpdough: could be a future enhancement if it's difficult to add.

cpdough (approved):
my comments don't need to block deployment. nice work dude.

also i didn't test these yet so someone else do that 😄

column_transforms/rasgo_concat/rasgo_concat.sql
```diff
@@ -0,0 +1,8 @@
+SELECT
+*
+, CONCAT(
+    {%- for obj in concat_list %}
+    {{obj}}{{ ", " if not loop.last else "" }}
+    {%- endfor %}
+)
```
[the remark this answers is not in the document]
szaghikian-rasgo: Snowflake automatically creates an alias column name based on the arguments passed to the function. For instance, the sample code in the YAML file:
```source.transform(
    source_code=s2,
    concat_list=["DS_WEATHER_ICON", "'some_str'", "'_5'"],
).preview_sql()```

produces an extra column named `CONCAT( DS_WEATHER_ICON, 'SOME_STR', '_5' )`. If this is cumbersome or crappy, I can just create an alias that we want using the arguments here.

File: column_transforms/rasgo_cast/rasgo_cast.sql [hunk 15]
```diff
@@ -0,0 +1,11 @@
+{% if col_list|length != type_list|length %}
+Rasgo Cast Error: The Column list must be the same length as the Type list.
```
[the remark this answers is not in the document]
szaghikian-rasgo: It does indeed. What u think?

szaghikian-rasgo:
@alexwolf22 you feel good merging this?

column_transforms/rasgo_cast/rasgo_cast.sql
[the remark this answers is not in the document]
cpdough: perfect

column_transforms/rasgo_concat/rasgo_concat.sql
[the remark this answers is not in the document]
cpdough: let's use cleanse_name and make the alias ourselves. i believe the snowflake column naming would have to be quoted which would break the rest of rasgo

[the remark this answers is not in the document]
szaghikian-rasgo: I added `cleanse_name` each place we create an alias. Lemme know what u think

PR #39, opened 2021-11-12
add rasgo_drop_columns UDT

cpdough:


column_transforms/rasgo_drop_columns/rasgo_drop_columns.yaml
```diff
@@ -0,0 +1,23 @@
...
+  exclude_cols:
+    type: column_list
+    description: A list of the columns from the dataset you want to drop. Any columns not in the exclude_cols list will be kept.
+example_code: |
+
+    t1 = source.transform(
+        transform_name='rasgo_drop_columns',
+        include_cols=["DS_WEATHER_ICON", "DS_DAILY_HIGH_TEMP", "DS_DAILY_LOW_TEMP"]
+    )
+
```
szaghikian-rasgo, line 22: For examples where it's ultra simple, do we want to showcase usage of all options? Like you could have here:
```
    t2 = source.transform(
        transform_name='rasgo_drop_columns',
        exclude_cols=["Some_column", "some other column"]
    )
```

column_transforms/rasgo_drop_columns/rasgo_drop_columns.sql
```diff
@@ -0,0 +1,44 @@
...
+    {%- else -%}
+        {%- set database, schema, table = rasgo_source_ref(source_id).split('.') -%}
+    {%- endif -%}
+        SELECT COLUMN_NAME FROM {{ database }}.information_schema.columns
+        WHERE TABLE_CATALOG = '{{ database }}'
+        AND   TABLE_SCHEMA = '{{ schema }}'
+        AND   TABLE_NAME = '{{ table }}'
+{%- endmacro -%}
+
+{% if include_cols and exclude_cols is defined %}
```
szaghikian-rasgo, line 18: does this need to be `if include_cols is defined and exclude_cols is defined`?

szaghikian-rasgo (approved):
Check the comment about `is defined` usage in the jinja template. Otherwise, LGTM

column_transforms/rasgo_drop_columns/rasgo_drop_columns.sql
szaghikian-rasgo: never mind. Looks like `is defined` is just a truthiness check so `if include_cols` == `if include_cols is defined`.

column_transforms/rasgo_drop_columns/rasgo_drop_columns.yaml
cpdough: good idea - will add

PR #40, opened 2021-11-15
Train test split enhancements

cpdough:
adding support for a random sampling in addition to the order by split

### DIFF
--- a/column_transforms/rasgo_math/rasgo_math.sql
+++ b/column_transforms/rasgo_math/rasgo_math.sql
@@ -1,0 +1,6 @@
+SELECT *
+{%- for math_op in math_ops %}
+    , {{math_op}} as {{cleanse_name(math_op)}}
+{%- endfor %}
+
+FROM {{source_table}}
--- a/column_transforms/rasgo_math/rasgo_math.yaml
+++ b/column_transforms/rasgo_math/rasgo_math.yaml
@@ -1,0 +1,18 @@
+name: rasgo_math
+transform_type: column
+source_code: rasgo_math.sql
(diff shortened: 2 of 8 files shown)


cpdough:
@griffatrasgo  it seems like the SQL on this one is reversed and it should be:
`CASE WHEN MOD(RANDOM(),  1/{{train_percent}}) = 0 THEN 'TEST' ELSE 'TRAIN' END AS TT_SPLIT`

can you take a look when you have time? thanks

griffatrasgo:
> `/{{train_percent}}) = 0 THEN 'TEST' ELSE 'TRAIN' END AS TT_SPLIT`
> 
> can you take a look when you have time? thanks

Yes, good catch, this code is correct with `1/{{train_percent}}) = 0`

Here's a test to confirm the results:
```WITH SPLITS AS (
SELECT *,
  CASE WHEN ROW_NUMBER() 
    OVER (ORDER BY DATE) < (COUNT(1) OVER () * .8) THEN 'TRAIN' ELSE 'TEST' END AS TT_SPLIT_ORDERED,
  CASE WHEN 
    MOD(RANDOM(),  1/.8) = 0 THEN 'TEST' ELSE 'TRAIN' END AS TT_SPLIT_RANDOM
FROM RASGOCOMMUNITY.PUBLIC.DARKSKY_DAILY_FEATURES
)

SELECT COUNT(TT_SPLIT_ORDERED), COUNT(TT_SPLIT_RANDOM) FROM SPLITS;```

griffatrasgo (approved):

PR #43, opened 2021-11-15
add cleanse_name to fix errors

cpdough:


cpdough:
@engelAnalytics can you try passing this one for your transform with `source_code` to confirm it resolves your issue?

PR #45, opened 2021-11-16
order and case jinja and docs

szaghikian-rasgo:
I added UDTs for a simple order statement and for a more expanded `CASE-WHEN` functionality. I added a few things to the `CASE` UDT I want y'all to review:
1. If a user passes in as the first object in the sub-list in a `replace_dict` a string that matches the name of a column in the data set, I have unquoted that string. This allows us to compare columnar values.
2. If a user passes just a string into the `replace_dict`, this is treated as the new column name resulting from the transformation. The default for this is `col_name_REPLACED`
3. If a user passes in a single value list to the list in the `replace_dict`, this is treated as the default value for the `CASE-WHEN` statements. The default for this is `NULL`


column_transforms/case/case.yaml
```diff
@@ -0,0 +1,28 @@
+name: case
```
alexwolf22, line 1: Make sure to add this UDt to the base `README.md` file

column_transforms/case/case.yaml
```diff
@@ -0,0 +1,28 @@
...
+  The new column generates values based on the criteria layed out in the nested list for that column. The first value is a value in the existing column for which we wish to provide a case. This value can be a column name in the data set, a fixed string, an integer, or None.
+
+  The second value is the desired output value in the new column. This can also be a column name in the data set, a fixed string, an integer, or None.
+
+  To set a default value for the new column, include a list that is one item long.
+
+  To set the output column name, include an unlisted string in the arguments.
+arguments:
+  replace_dict:
+    type: replace_dict
```
cpdough, line 16: we connected on this directly

PR #46, opened 2021-11-16
Added standard_scaler

cpdough:
@engelAnalytics can you check this one out?


cpdough:
@engelAnalytics I implemented the approach we discussed about allowing for an override on the calculated values with static values. check it out and see what you think.

rest of team for visibility to this approach.

column_transforms/standard_scaler/standard_scaler.yaml
```diff
@@ -0,0 +1,33 @@
+name: standard_scaler
+transform_type: column
+source_code: standard_scaler.sql
```
alexwolf22, line 3: reminder arent keeping these fields in YAML

PR #49, opened 2021-12-02
first wave of changes

cpdough:
@phillip-rasgo  these are merge-able without syntax changes; the others will require a bit more effort. will put those in a draft pr until we're ready to cut over

### DIFF
--- a/column_transforms/datediff/datediff.yaml
+++ b/column_transforms/datediff/datediff.yaml
@@ -2,14 +2,14 @@ name: datediff
 description: Calculates the difference between two date, time, or timestamp expressions based on the date or time part requested. The function returns the result of subtracting the second argument from the third argument.
 arguments:
   date_part:
-    type: string
+    type: date_part
     description: |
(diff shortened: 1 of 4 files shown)


column_transforms/datediff/datediff.yaml
```diff
@@ -2,14 +2,14 @@ name: datediff
 description: Calculates the difference between two date, time, or timestamp expressions based on the date or time part requested. The function returns the result of subtracting the second argument from the third argument.
 arguments:
   date_part:
-    type: string
+    type: date_part
     description: |
        Must be one of the values listed in [Supported Date and Time Parts](https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts)
   date_val_1:
-    type: string
+    type: mixed_value
```
phillip-rasgo, line 9: Are we OK with this just being a text field for the non-column input?

phillip-rasgo (approved):

column_transforms/datediff/datediff.yaml
cpdough: for now, i think yeah, we can just specify in docs the format

PR #57, opened 2021-12-08
Remove rasgo source ref

szaghikian-rasgo:
Removed all instances of `rasgo_source_ref`. Changed the `get_source_col_names` macro to only take an FQTN, which it now does everywhere it's called

### DIFF
--- a/table_transforms/simple_join/simple_join.yaml
+++ b/table_transforms/simple_join/simple_join.yaml
@@ -4,7 +4,7 @@ description: |
 
 arguments:
   join_table:
-    type: dataset
+    type: table
     description: Dataset object to join with the source_table
   join_type:
     type: string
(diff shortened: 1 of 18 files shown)


table_transforms/simple_join/simple_join.yaml
```diff
@@ -4,20 +4,21 @@ description: |
 
 arguments:
   join_table:
-    type: source
-    description: source ID of the table to join with the source_table
+    type: dataset
```
phillip-rasgo, line 7: We do not want to restrict joins to only working on datasets rather than operation tables.  Shouldn't this be a table?

cpdough: just made the change to 'table'

cpdough (approved):
arg type change is in to mention table instead of dataset. should be good now

PR #58, opened 2021-12-10
Dont use dumb lists

cpdough:
trying to remove the ordinal list matching stuff since it is confusing. dicts make it easier and better for the UI i think. unless @phillip-rasgo  says i'm wrong.

did i miss any transforms that use ordinal list matching?


column_transforms/datepart/datepart.sql
```diff
@@ -1,3 +1,9 @@
 SELECT *,
 {% for col in date_columns %}DATE_PART('{{date_part}}', {{col}}) as {{col}}_datepart_{{date_part}} {{ ", " if not loop.last else "" }}{% endfor %}
+from {{ source_table }}
+
+SELECT *,
+{%- for target_col, date_part in dates.items() %}
+    DATE({{target_col}}, '{{date_part}}') as {{target_col}}_date {{ ", " if not loop.last else "" }}
+{%- endfor %}
```
szaghikian-rasgo, line 8: wait why are there two select statements here? Or are you planning to delete the first in favor of the second?

szaghikian-rasgo: Actually I think that you may have dropped a date cast into this template?

column_transforms/new_cast/new_cast.yaml
```diff
@@ -0,0 +1,20 @@
+name: new_cast
+description: |
+  Cast selected columns to a new type
+arguments:
+  casts:
+    type: cast_dict
```
szaghikian-rasgo, line 6: do we need a new type here? if we're gonna start using k-v more thoroughly in UDTs, should we just have a dict type?

column_transforms/datepart/datepart.sql
cpdough: thanks for the find

column_transforms/new_cast/new_cast.yaml
cpdough: good call i forgot about the new column_value_dict just for these types of things

column_transforms/datepart/datepart.yaml
```diff
@@ -1,19 +1,20 @@
...
+  An exhaustive list of valid date parts can be [found here](https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts).
 arguments:
-  date_columns:
-    type: column_list
-    description: names of column(s) you want to date-part
-  date_part:
-    type: date_part
-    description: the desired grain of the date; an exhaustive list of valid date parts can be [found here](https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts).
+  dates:
+    type: datepart_dict
```
phillip-rasgo, line 8: This type `datepart_dict` is not yet supported on the UI side.   Should we add it?

docs/column_transforms/new_cast.md
```diff
@@ -0,0 +1,36 @@
...
+# new_cast
+
+Cast selected columns to a new type
+
+
+## Parameters
+
+| Argument |   Type    |                                    Description                                     |
+| -------- | --------- | ---------------------------------------------------------------------------------- |
+| casts    | cast_dict | A dict where the keys are columns and the values are the new type to cast them to. |
```
phillip-rasgo, line 12: I'm not seeing the corresponding YAML for this, but this look like a new type as well.  (I do think both of these new types are a good idea to add.)

PR #62, opened 2021-12-14
No more sources

cpdough:
this tiny PR removes all mention of sources from code samples and other docs components.

it also cleans up a few argument names that were not as clear to the user as they could be.


bstempi (approved):
Such a tiny PR 😄 .

I did a search on "source" within the PR to see what changed and so far it looks good to me. Obviously, I did not go looking through our code to find inappropriate uses of "source," but rather agree that these changes remove undesired uses.

szaghikian-rasgo (approved):
"tiny"

PR #63, opened 2021-12-14
add multi_join

cpdough:
adding a multi_join. @cnash6 what would pyrasgo do with my 'table_list' arg type? would it inject FQTNs for all those?

### DIFF
--- a/docs/table_transforms/multi_join.md
+++ b/docs/table_transforms/multi_join.md
@@ -1,0 +1,37 @@
+
+
+# multi_join
+
+Join n number of datasets with the 'base' dataset, using a consistent join_type and consistent list of join_columns across all joins.
+
+
+## Parameters
+
+|   Argument   |    Type     |                                              Description                                              |
+| ------------ | ----------- | ----------------------------------------------------------------------------------------------------- |
+| join_tables  | table_list  | Datasets to join with the source_table                                                                |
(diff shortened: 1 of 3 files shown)


docs/table_transforms/multi_join.md
```diff
@@ -0,0 +1,37 @@
...
+Join n number of datasets with the 'base' dataset, using a consistent join_type and consistent list of join_columns across all joins.
+
+
+## Parameters
+
+|   Argument   |    Type     |                                              Description                                              |
+| ------------ | ----------- | ----------------------------------------------------------------------------------------------------- |
+| join_tables  | table_list  | Datasets to join with the source_table                                                                |
+| join_type    | join_type   | Type of join to run against the base dataset (either LEFT, RIGHT, or INNER)                           |
+| join_columns | column_list | Columns to join on. Can be one or more columns but must be named the same thing between all datasets. |
```
bstempi, line 14: Is this because we expect an end-user to do column renames before getting to this point?

szaghikian-rasgo (approved):
Seems legit

docs/table_transforms/multi_join.md
cpdough: yes, if they want to join N number of tables... so rather than try to support N-1 number of different join equalities passed through a dict, we just apply an artificial limitation of consistent column names to minimize the surface area of possible failures. if they want to be fancy, they use the standard join which only supports joining 2 tables

PR #65, opened 2021-12-17
create or update transforms

alexwolf22:
Update the script to update a transform if one with that’s a name already exists. If not create it.


szaghikian-rasgo (approved):
I'm assuming you already checked for this, but my only concern would be any other consumers of `get_all_rasgo_transform_name_and_ids`, which expect to get a set returned, but are now instead getting a dict. Otherwise, looks pretty g to me

PR #66, opened 2021-12-18
Multi union

cpdough:
Supports unioning across n number of tables, given a list of common column names between each of them


szaghikian-rasgo (approved):
LGTM, very exciting

PR #68, opened 2021-12-20
delete transform in scipt

alexwolf22:
- Delete any transforms in Rasgo which aren't in the UDT repo 
- - rename script from `create_and_update_transforms.py` to `publish_transforms.py`


szaghikian-rasgo (approved):
I like

szaghikian-rasgo:
Any risk of this overwriting public transformations or private transformations not written by us?

alexwolf22:
@szaghikian-rasgo No because we're going to run the script as community organization, so could only delete community ones.

We have backend API validation in prices to make sure users have access to update/delete transforms too.

PR #73, opened 2021-12-22
Add Target encode

cpdough:
adding a target encode transform for feature engineering.

@engelAnalytics can you verify that this gives the same results as sci-kit learn target encoder? want to make sure i have the logic right.


engelAnalytics:
This returns the same value as the scikit-learn version with the caveat that this only returns three digits after the decimal, while scikit-learn does not round. Probably doesn't matter for this transform.

PR #78, opened 2022-01-15
market_basket

cpdough:
This is a transform to make transactional market basket analysis really easy. Test it out with:

`internet_sales = rasgo.get.dataset(74)

ds2 = internet_sales.market_basket(transaction_id='SALESORDERNUMBER',
              agg_column='PRODUCTKEY',
              sep='|')`

### DIFF
--- a/table_transforms/market_basket/market_basket.yaml
+++ b/table_transforms/market_basket/market_basket.yaml
@@ -6,7 +6,7 @@ description: |
 arguments:
   transaction_id:
     type: column
-    description: Column to group multiple line items into a single transaction
+    description: Column identifying a unique event ID (i.e., transaction) for which to aggregate line items
   sep:
     type: value
     description: Text separator to use when aggregating the strings, i.e. ', ' or '|'.
--- a/table_transforms/market_basket/market_basket.yaml
+++ b/table_transforms/market_basket/market_basket.yaml
@@ -1,0 +1,22 @@
+name: market_basket
+description: |
+  Analyze historical transaction contents to understand products that are frequently purchased together.
+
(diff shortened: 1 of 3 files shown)


table_transforms/market_basket/market_basket.yaml
```diff
@@ -0,0 +1,22 @@
+name: market_basket
+description: |
+  Analyze historical transaction contents to understand products that are frequently purchased together.
+
+  This approach uses a transactional table to aggregate each product purchased in a transaction, and then aggregates transactions together to look for common patterns.
+arguments:
+  transaction_id:
```
szaghikian-rasgo, line 7: Why is this name Transaction ID? Could we name it something more meaningful, or do you think product analysts will get the idea? Maybe I'm just unhappy with the description. I think something like `Name of column identifying a unique transaction for which to count` works better?

table_transforms/market_basket/market_basket.yaml
```diff
@@ -0,0 +1,22 @@
...
+    type: column
+    description: Column to group multiple line items into a single transaction
+  sep:
+    type: value
+    description: Text separator to use when aggregating the strings, i.e. ', ' or '|'.
+  agg_column:
+    type: column
+    description: Product ID or description to use when aggregating into transactions
```
szaghikian-rasgo, line 15: I think this var name & description suffers from the same complaint above. My personal opinion, I'm happy to hear counter arguments

szaghikian-rasgo (commented):
My complaints are mostly around the template variable names and their descriptions. I do like the example code, thought - it does clarify some of the info implicit in the current descriptions. I'm just not sure users will get there by themselves

table_transforms/market_basket/market_basket.yaml
cpdough: done

PR #79, opened 2022-01-16
Lag aggregate

cpdough:
This one is cool. the code sample can be used to test it easily. curious what Professor @engelAnalytics thinks, if this is usable as is for time series forecasting.


table_transforms/lag_aggregate/lag_aggregate.yaml
```diff
@@ -0,0 +1,36 @@
...
+  Temporally aggregates a continuous column (i.e. sales or quantity) over one or more time windows, producing a column per window.
+
+  Best used in time series forecasting or analysis when you need to analyze performance over a variety of historical periods.
+arguments:
+  group_by:
+    type: column_list
+    description: Column(s) to group by when aggregating the target number
+  windows:
+    type: value_list
+    description: List of numeric intervals to use when calculating the lag windows for the aggregation. One column will be created for each window value in the list.
```
szaghikian-rasgo, line 12: I think we should make it clearer that this window value is for the "list of numeric intervals *of the date column*"

szaghikian-rasgo (approved):
Code and docs look good. I'll test functionality and then finalize review in the next day or so

szaghikian-rasgo:
My approval was supposed to be a PR comment - IGNORE THE GREEN CHECK

table_transforms/lag_aggregate/lag_aggregate.yaml
cpdough: changed

PR #82, opened 2022-01-17
Add Linear Regression transform

cpdough:
new transform. check it out. code sample works to test it.


table_transforms/linear_reg/linear_reg.yaml
```diff
@@ -0,0 +1,28 @@
+name: linear_reg
+description: |
+  Fit a simple linear regression and return the formula. Optionally, use one or more group_by columns to create a regression per unique grouping.
+  
+  Currently, only supports a single independent variable.
+arguments:
+  group_by:
+    type: column_list
+    description: Optional, columns to group by before building the linear regression model. Use this field to create multiple models (one per unique grouping)
```
szaghikian-rasgo, line 9: Are you setting an optional variable the right way here?

szaghikian-rasgo (approved):
Looks good to me, but I had a question about making variables optional using our YAML files

table_transforms/linear_reg/linear_reg.yaml
cpdough: good call thanks

PR #83, opened 2022-01-17
Add is optional flag to udt args

alexwolf22:
This PR adds to the UDT YAML files a new flag `is_optional`, which is set to true for every optional UDT arg in this repo. it doesn't need to bet set in the YAML for UDT args which are required. 

We are also now adding a new column to the UDT doc that's generated indicating if a UDT arg is optional. 

**NOTE**: All the backend and Pyrasgo work is already in place/merged/live to store these in the `source_transform_argument` Postgres table in the column `is_optional`. Once merged with master the git workflow will update our prod DB ASAP.

## Important Ask for Reviewers
I had to manually go through each UDT jinja template/YAML file to check if the arg is optionally or not. I am pretty sure I got all of them but can you please
1. Make sure the ones I labeled as optional are in fact (pretty sure they are)
2. Let me know if I am missing any UDT args which are optional but weren't marked in this PR.

### DIFF
--- a/python/utils.py
+++ b/python/utils.py
@@ -121,7 +121,8 @@ def get_table_values(transform_args: Dict) -> List[List[str]]:
         row_data = [
             arg_name,
             arg_info['type'],
-            arg_info['description']
+            arg_info['description'],
+            arg_info.get('is_optional', '')
         ]
         all_data.append(row_data)
     return all_data
--- a/column_transforms/drop_columns/drop_columns.yaml
+++ b/column_transforms/drop_columns/drop_columns.yaml
@@ -8,9 +8,11 @@ arguments:
   include_cols:
     type: column_list
     description: A list of the columns from the dataset you want to keep.
+    is_optional: true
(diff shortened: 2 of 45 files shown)


python/utils.py
```diff
@@ -121,7 +121,8 @@ def get_table_values(transform_args: Dict) -> List[List[str]]:
         row_data = [
             arg_name,
             arg_info['type'],
-            arg_info['description']
+            arg_info['description'],
+            arg_info.get('is_optional', '')
```
bstempi, line 125: Why default to `''` instead of `FALSE`?

bstempi (approved):
LGTM

cpdough:
there are some .md files in here that do not have companion YAML file changes in the PR, such as math.md ... is this accidental?

alexwolf22:
@cpdough No, it's not. The table describing the Transform Args now has a new column called `Is Optional` in the docs, and was added to every transform markdown file.

PR #87, opened 2022-01-19
Does this have side effects?

sam-rasgo:
I have a UI PR open that updates the arguments we display for a cast transform in the UI. I built it with the assumption that I could change the argument passed in for cast from "column_value_dict" to "cast_value_dict". Also, I made the assumption that pyrasgo could handle this change because it would only change how the transform was shown in the SDK, and not change anything about how it was stored.

I'm raising this PR to check these assumptions. Let me know your thoughts/if I'm wrong

Here's the UI PR: https://github.com/rasgointelligence/RasgoUI/pull/206

### DIFF
--- a/column_transforms/cast/cast.yaml
+++ b/column_transforms/cast/cast.yaml
@@ -3,7 +3,7 @@ description: |
   Cast selected columns to a new type
 arguments:
   casts:
-    type: column_value_dict
+    type: cast_value_dict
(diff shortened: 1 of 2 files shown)


cpdough (approved):
sweet - nice work @sam-rasgo ... arg type is only used by the UI so no unintended side effects here 👍

sam-rasgo:
@cpdough I haven't worked with this repo yet. What are the steps after merging to get this live?

PR #94, opened 2022-01-24
Update cast.sql

cpdough:


griffatrasgo (approved):
I'm sure we're about to get 10 new error message, but we won't know what they are until we merge :D

PR #96, opened 2022-01-31
Fix join aliasing

cpdough:
this PR changes how 'join' handles columns that have the same name in both tables


griffatrasgo (approved):
This logic makes sense

PR #100, opened 2022-02-02
Window aggregates

griffatrasgo:
Here's a stab in the dark to satisfy [RAS-2397](https://linear.app/rasgoml/issue/RAS-2397/rolling)

The mechanics of rolling vs sliding windows are different enough that I broke them into separate transforms. If we want to combine these and support a whole buncho toggles, we can discuss.

Still need to test er


engelAnalytics:
I think the rolling aggregation is closer to what pandas calls resample: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.resample.html
Than what pandas calls rolling: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html

For rolling, if I ask for a rolling window with length week, for each row, we need to aggregate over the prior week from that row. So for a week aggregation first row at 2022-01-31 at 10:10, aggregate between that date and 2022-01-24 at 10:10.

If next row is at 2022-01-31 at 10:20, the aggregation should be between that new date and 2022-01-24 at 10:20.

PR #102, opened 2022-02-04
Fancy datespine

griffatrasgo:
This is the first bullet point in [RAS-2546](https://linear.app/rasgoml/issue/RAS-2546/3-timeseries-prep-transforms)

It allows users to apply a datespine to groups within a table, instead of to the entire table.

### DIFF
--- a/docs/table_transforms/datespine.md
+++ b/docs/table_transforms/datespine.md
@@ -4,8 +4,11 @@
 
 This transform generates a date spine for your date index, which can replace your date index column for modeling.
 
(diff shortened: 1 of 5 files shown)


cpdough:
i'm leaning toward combining these into a single `datespine` transform with optional group_by.... curious if @itsamejoshab agrees with that from a user standpoint

if so, i'm happy to work on merging the templates

itsamejoshab:
- created a dataset for testing these functions: `rasgo.get.dataset(2655)`
- made some fixes to `datespine_groups` - checks out OK!
- still need to do more testing

itsamejoshab (approved):

itsamejoshab:
its so fancy. tested it with dataset rasgo.get.dataset(2655)

PR #106, opened 2022-02-08
Changes to support pypi packaging

griffatrasgo:
The Rasgo open source package will support non-Snowflake DataWarehouses much sooner than our API will. Two major changes will be needed to support running these transforms in an open source package:
1. Don't assume source_table arg will conform to Snowflake's fqtn format
2. Don't assume we can hit a view named `information_schema` in the DW

This PR does not address these issues entirely, but chips away at them by:
- removing calls to information_schema when a simple SELECT TOP 0 will suffice
- refactors macros that use .split() with an assumption that the table will be a fqtn

...Needs more testing...


griffatrasgo:
snuck all these changes in my other PR to test them in RQL before merging

PR #107, opened 2022-02-08
Preparing for pypi packaging

griffatrasgo:
This PR prepares our repo to become an open-source package on pypi.
Changes:
- move all package files under folders: /rasgoudt/rasgoudt
- add a function in the /main.py file to serve these to python consumers
- add pypi files: DESC, setup.py, version
- add docker resources to assist in deployment

Updates complements of @alexwolf22:
- edit Rasgo internal publishing scripts to use new folder paths

TODO:
- test
- add code to auto-increment package version on deploy
- confirm package name
- publish package


bstempi (approved):
Holy PR, Batman! 

The description file and main.py LGTM.

rasgoudt/requirements.txt
```diff
@@ -0,0 +1 @@
+pyyaml
```
RasgoDerek, line 1: You don't need a requirements.txt and a setup.py, it just creates two places where people can screw up adding package dependencies (like I did)

I recommend removing this file and using `pip install .` if you want to install the package locally during testing and whatnot

RasgoDerek (changes requested):

rasgoudt/setup.py
```diff
@@ -0,0 +1,43 @@
...
+
+version = {}
+with open(os.path.join(_here, 'rasgoudt', 'version.py')) as f:
+    exec(f.read(), version)
+
+with open(os.path.join(_here, 'DESCRIPTION.md'), encoding='utf-8') as f:
+    long_description = f.read()
+
+setup(
+    name='rasgoudt',
```
cnash6, line 14: not `rasgo-udt`?

rasgoudt/requirements.txt
griffatrasgo: we use the `requirements.txt` filte to import dependencies in the docker container while building/publishing the dist. We may not want to keep the docker deployment process here. I kept it bc that's what we do in pyrasgo and figured we should debate a change as a team. I'll bring this topic to standup PL tomorrow.

rasgoudt/setup.py
griffatrasgo: why not `christian-eats-worms`?

griffatrasgo:
>     * leave out the RQL use case and focus purely on `serve_rasgo_transform_templates` so that we don't distract from the _purity_ here and so that we don't have to maintain demo code in both places.  Totally on board for shilling and linking to the other repos/sites though.

Guessing this is referring to the readme? Good catch. Looks like this was all copied from the docs homepage. Makes sense that the readme should be de-coupled from the pyrasgo code.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN chmod +x /usr/bin/entrypoint
+ENTRYPOINT ["/usr/bin/entrypoint"]
+
+USER ${PRIMARY_USER}
+WORKDIR ${HOME_DIR}
+
+
+FROM pythonapp AS rasgoudt
+ARG HOME_DIR=/var/lib/rasgo
+COPY rasgoudt/requirements.txt ${HOME_DIR}
```
alexwolf22, line 29: do we maybe want the root dir in the container to be called `rasgo_transforms` instead of `rasgoudt` to keep it consistent with Github repo name?

rasgoudt/rasgoudt/column_transforms/drop_columns/drop_columns.sql
```diff
@@ -3,14 +3,7 @@ Jinja Macro to generate a query that would get all
...
 {%- macro get_source_col_names(source_table_fqtn=None) -%}
-    {%- set database, schema, table = '', '', '' -%}
-    {%- if source_table_fqtn -%}
-        {%- set database, schema, table = source_table_fqtn.split('.') -%}
-    {%- endif -%}
-        SELECT COLUMN_NAME FROM {{ database }}.information_schema.columns
-        WHERE TABLE_CATALOG = '{{ database|upper }}'
-        AND   TABLE_SCHEMA = '{{ schema|upper }}'
-        AND   TABLE_NAME = '{{ table|upper }}'
+    select top 0 * from {{ source_table_fqtn }}
```
alexwolf22, line 6: Did we retest these updated transforms with the source_col_name changes?

rasgoudt/requirements.txt
griffatrasgo: per parkinglot, we now import the reqs in the setup.py file from this file so we'll only need to make the change in one place :D

Dockerfile
griffatrasgo: yah, good call

PR #110, opened 2022-02-12
Replace CTEs with Subqueries

griffatrasgo:
Currently rasgoql's CTE chain logic doesn't play nice with CTEs in transforms. That will need to be fixed eventually. 
The few transforms that use CTEs currently can all be switched to subqueries without performance impact. This can buy us some time.


griffatrasgo:
Fixed this in RQL instead of here

PR #112, opened 2022-02-16
Change top 0 to limit statements

griffatrasgo:
Ansii compliance change. These should all be functionally the same


szaghikian-rasgo (approved):
Yep, LGTM. Got me thinking we should do something similar with our `TOP` queries in the backend

PR #113, opened 2022-02-16
Support additional DBs

griffatrasgo:
This is the first step towards running transforms on non-Snowflake dbs.

Changes included here:
- accept a `datawarehouse` param to look for override files
- serve override templates if a db override file is found

TODO in future PR:
- comb through each transform and filter them into ansii-compliant vs override files
- update internal publishing script to only pull Snowflake files for webapp


rasgotransforms/rasgotransforms/main.py
```diff
@@ -111,18 +115,26 @@ def get_root_dir() -> Path:
...
     """
-    From a transform name and type load and return it's source code as a string
+    Return a transform's source code as a string
     """
     root_dir = get_root_dir()
     source_code_path = root_dir / f"{transform_type}_transforms" / transform_name / f"{transform_name}.sql"
-    fp = open(source_code_path)
-    source_code = fp.read()
-    fp.close()
+    source_code_override_path = root_dir / f"{transform_type}_transforms" / transform_name/ datawarehouse / f"{transform_name}.sql"
```
alexwolf22, line 128: We should maybe raise an error if the `datawarehouse` is none here or make it a required param.

Potentially make an enum value for data warehouse as well?

rasgotransforms/rasgotransforms/main.py
```diff
@@ -77,31 +82,30 @@ def serve_rasgo_transform_templates() -> List[TransformTemplate]:
             )
     return template_list
 
-def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
+def load_all_yaml_files(
+        datawarehouse: str = None
```
alexwolf22, line 86: Same here about having `datawarehouse` here be required param/not None or make a Enum

rasgotransforms/rasgotransforms/main.py
```diff
@@ -77,31 +82,30 @@ def serve_rasgo_transform_templates() -> List[TransformTemplate]:
...
+    ) -> Dict[str, Dict[str, Dict]]:
     """
-    Load and return all the yaml files in the dirs <root>/<transform_type>_transforms
-    If new transform type/dir added be sure to add above in List TRANSFORM_TYPES
+    Load and return all the yaml files in the dir <root>/<transform_type>_transforms
```
alexwolf22, line 89: Could be helpful to add a quick note in the docstring about returning the data warehouse jinja if exists then the default

alexwolf22 (commented):
Great job on this! Really like the structure of the datawarehouse Jinja organization.

One thing I just want to address is the `None` potential value for data warehouse func param.

Also, aren't we going to need to change the `publish_transorms` GitHub action to be compliant with only the snowflake transforms?

rasgotransforms/rasgotransforms/main.py
griffatrasgo: Good call. I'll make it required

rasgotransforms/rasgotransforms/main.py
griffatrasgo: good call on the enum

griffatrasgo:
> Also, aren't we going to need to change the publish_transorms GitHub action to be compliant with only the snowflake transforms?

Yes, I should include that in this PR vs pushing to a future one. Will make the change.

python/utils.py
```diff
@@ -31,9 +31,11 @@ def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
         # Get list of all transform of certain type
         transform_names = [x.name for x in transform_type_dir_path.rglob("*/**")]
         for transform_name in transform_names:
-            transform_dir = transform_type_dir_path / transform_name
-            transform_yaml_path = transform_dir / f"{transform_name}.yaml"
+            transform_yaml_path = transform_type_dir_path / transform_name / f"{transform_name}.yaml"
+            transform_yaml_override_path = transform_type_dir_path / transform_name / constants.RASGO_DATAWAREHOUSE / f"{transform_name}.yaml"
 
+            if Path(transform_yaml_override_path).exists():
```
alexwolf22, line 37: `transform_yaml_override_path` is already a path obj I'm pretty sure, so no need to wrap it in a path obj again

python/utils.py
```diff
@@ -111,9 +113,11 @@ def get_transform_source_code(transform_type: str, transform_name: str) -> str:
     """
     transform_type_dir = _get_udt_repo_dir() / f"{transform_type}_transforms"
     source_code_path = transform_type_dir / transform_name / f"{transform_name}.sql"
-    fp = open(source_code_path)
-    source_code = fp.read()
-    fp.close()
+    source_code_override_path = transform_type_dir / transform_name / constants.RASGO_DATAWAREHOUSE / f"{transform_name}.sql"
+    if Path(source_code_override_path).exists():
```
alexwolf22, line 117: Same here about wrapping with path

docs/column_transforms/bin.md
```diff
@@ -2,6 +2,7 @@
 
 # bin
 
+(Snowflake-specific transform)
```
alexwolf22, line 5: Just curious why did we add this comment here? Is that because it's only possible in snowflake?

Would this cause potential problems for people using our library?

rasgotransforms/rasgotransforms/main.py
```diff
@@ -77,52 +91,71 @@ def serve_rasgo_transform_templates() -> List[TransformTemplate]:
...
 
-        # Get list of all transform of certain type
         transform_names = [x.name for x in transform_type_dir_path.rglob("*/**")]
         for transform_name in transform_names:
-            transform_dir = transform_type_dir_path / transform_name
-            transform_yaml_path = transform_dir / f"{transform_name}.yaml"
+            transform_yaml_path = transform_type_dir_path / transform_name / f"{transform_name}.yaml"
+            transform_yaml_override_path = transform_type_dir_path / transform_name / datawarehouse / f"{transform_name}.yaml"
+
+            if Path(transform_yaml_override_path).exists():
```
alexwolf22, line 145: same hejre with Path being redefined

rasgotransforms/rasgotransforms/main.py
```diff
@@ -77,52 +91,71 @@ def serve_rasgo_transform_templates() -> List[TransformTemplate]:
             )
     return template_list
 
-def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
+def _check_datawarehouse(
+        input_value: str
+    ):
```
alexwolf22, line 96: Missing return type annotation

rasgotransforms/rasgotransforms/main.py
```diff
@@ -77,52 +91,71 @@ def serve_rasgo_transform_templates() -> List[TransformTemplate]:
...
+    supported_dws = "'"+"', '".join([e.value for e in Datawarehouse])+"'"
+    try:
+        Datawarehouse[input_value.upper()]
+    except Exception:
+        raise ValueError(f'datawarehouse parameter accepts values: {supported_dws}')
```
alexwolf22, line 101: Nice User expereince here!

alexwolf22 (approved):
Looks good! Made some minor last comments

docs/column_transforms/bin.md
griffatrasgo: this was just a test file. removing it

PR #120, opened 2022-02-25
Alex/publish transforms with tags

alexwolf22:
## Context

**Linear Issue:** [Add “tags” to transform yaml file and data model so that we can better organize them](https://linear.app/rasgoml/issue/RAS-2487/add-tags-to-transform-yaml-file-and-data-model-so-that-we-can-better)

**DR Champion:** @cpdough 

## Summary of changes
This PR allows people to at the top level of a transform YAML file, specify `tags` which can be a string or a list of strings.

The auto-transform deploy script will then incorporate these tags from yaml when making the PyRasgo calls to create a transform.

### Testing
Tested changing tag types in yaml and making sure it is updated properly in the db/returned in PyRasgo

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,3 +1,3 @@
 pytablewriter
-pyrasgo>=0.4.3
+pyrasgo>=0.4.27
 pyyaml
--- a/python/utils.py
+++ b/python/utils.py
@@ -28,15 +28,15 @@ def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
     for transform_type in constants.TRANSFORM_TYPES:
         transform_type_dir_path = _get_udt_repo_dir() / f"{transform_type}_transforms"
 
-        # Get list of all transform of this type,
-        # by getting names of sub-directory one level down
+        # Get list of all transform names of this type,
+        # by looking at sub-directory names one level down
         transform_names = [
             x.name for x in transform_type_dir_path.iterdir() if x.is_dir()
         ]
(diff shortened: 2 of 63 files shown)


python/utils.py
```diff
@@ -28,14 +28,19 @@ def load_all_yaml_files() -> Dict[str, Dict[str, Dict]]:
     for transform_type in constants.TRANSFORM_TYPES:
         transform_type_dir_path = _get_udt_repo_dir() / f"{transform_type}_transforms"
 
-        # Get list of all transform of certain type
-        transform_names = [x.name for x in transform_type_dir_path.rglob("*/**")]
+        # Get list of all transform names of this type,
+        # by looking at sub-directory names one level down
+        transform_names = [
+            x.name for x in transform_type_dir_path.iterdir() if x.is_dir()
+        ]
```
alexwolf22, line 35: @griffatrasgo FYI, we may want to change the `RasgoQL` packages to get transform names of a certain type this way as well.

Doing `transform_type_dir_path.rglob("*/**")` would return all subdirectories recursively, so we were trying to load YAML files that didn't exist causing the exception to happen.

python/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pytablewriter
-pyrasgo>=0.4.3
+pyrasgo>=0.4.27
```
szaghikian-rasgo, line 2: Why the version change?

szaghikian-rasgo (approved):
Looks good

PR #122, opened 2022-02-28
add data samples

szaghikian-rasgo:
@alexwolf22 Check it - thinking about adding data samples to the YAML files in this repo.


cpdough:
i think we could further simplify and improve the data samples by:
- not showing too many extra columns in the pre-transformed data
- making sure that the rows match between the pre- and post-, when possible... AKA making sure that ticker values that show up as columns in the post-transformed are present in the pre-transformed sample

szaghikian-rasgo:
Was thinking the same thing for the first point. I think I'm going to only include columns that are pertinent to the transforms. 

For the second point - for this pivot example, in the pre-transform data, I'm probably going to filter where ticker is in the values we pivot out, and then order by date. In the post-transform data, I'm going to order by date so that we have at least some rows that match

python/doc_generator.py
```diff
@@ -47,7 +54,12 @@ def _get_transform_markdown(transform_yaml: Dict,
         md.h2('Parameters'),
         md.table(transform_yaml['arguments']),
         md.h2('Example'),
+        md.text(DATASET_PREVIEW_TEXT),
```
alexwolf22, line 57: I feel like an header (`H2` `H3` or `H4`)  may be more appropriate for sectioning the documentation over just a text

python/doc_generator.py
```diff
@@ -47,7 +54,12 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+        md.python_code(DATASET_PREVIEW_CODE),
+        md.text(transform_yaml['preview-data']),
+        md.text(TRANSFORMED_DATA_PREVIEW_TEXT),
         md.python_code(transform_yaml['example_code']),
+        md.text(transform_yaml['post-transform-data']),
```
alexwolf22, line 62: So I don't know exactly what you're using to generate the markdown text for the tables, but we have a library installed already to make them for the Transform args. See the function `md.table`. I think it would be very tedious to do that manually.

rasgotransforms/rasgotransforms/table_transforms/pivot/pivot.yaml
```diff
@@ -19,9 +23,33 @@ arguments:
...
+  |  1 | 2018-09-25 |  8.26   | 1184.65  | 111.394  | 15.494  |
+  |  2 | 2014-06-20 |         |  554.837 |  76.2206 |         |
+  |  3 | 2018-07-31 | 17.5    | 1217.26  | 111.325  | 15.8779 |
+  |  4 | 2015-05-06 |         |  524.22  | 102.26   |         |
+  |  5 | 2017-08-02 |  7.2695 |  930.39  | 104.839  | 15.1934 |
+  |  6 | 2012-06-08 |         |  289.141 |  41.4007 |         |
+  |  7 | 2018-07-05 | 19.03   | 1124.27  | 102.444  | 16.0897 |
+  |  8 | 2009-12-24 |         |  308.085 |  28.1256 |         |
+  |  9 | 2008-09-03 |         |  231.338 |  27.5809 |         |
 example_code: |
```
alexwolf22, line 52: Feel like the example code should go above the data frames here, as the tables could be pretty long, and the user may not see the `example_code` section

rasgotransforms/rasgotransforms/table_transforms/pivot/pivot.yaml
```diff
@@ -19,9 +23,33 @@ arguments:
...
+  |  0 | 2017-11-14 | 21.5273 | 1026     |  99.5327 | 16.2433 |
+  |  1 | 2018-09-25 |  8.26   | 1184.65  | 111.394  | 15.494  |
+  |  2 | 2014-06-20 |         |  554.837 |  76.2206 |         |
+  |  3 | 2018-07-31 | 17.5    | 1217.26  | 111.325  | 15.8779 |
+  |  4 | 2015-05-06 |         |  524.22  | 102.26   |         |
+  |  5 | 2017-08-02 |  7.2695 |  930.39  | 104.839  | 15.1934 |
+  |  6 | 2012-06-08 |         |  289.141 |  41.4007 |         |
+  |  7 | 2018-07-05 | 19.03   | 1124.27  | 102.444  | 16.0897 |
+  |  8 | 2009-12-24 |         |  308.085 |  28.1256 |         |
+  |  9 | 2008-09-03 |         |  231.338 |  27.5809 |         |
```
alexwolf22, line 51: It seems super inefficient for us to manually generate these markdown strings for every transform each time we want to make changes.

Was thinking, maybe instead we could enter parameters for `preview-data`
- community dataset id to use
- columns to display int he output table

And then generate python code dynamically which will load that rasgo dataset into a df, and use the markdown library for writing the table in amrkdown.

For the `post-transform-data`. We could take as input in the yaml
- transform arg values to apply

And then do the same thing and generate the table dynamically.
`

alexwolf22 (changes requested):
Overall looks good!

My biggest request is that we try to automate, making the tables instead of manually creating them ( see my comment talking about a potential solution for this).

I agree with Patrick's two points/comments as well.

Another important thing, do we want to show multiple different kinds of outputs when the args change when executing the transform?

For example, in the pivot transform the output would be different for `list_of_vals` if supplied or not, and would be helpful for the user to see output examples and when it's used and not to understand how the transform works.

Pandas does a terrific job at this, and has with extremely simple input tables (maybe we want to do this). For example look at their doc showing how the [`pivot` function works](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pivot.html).

python/doc_generator.py
szaghikian-rasgo: The `MarkdownTableWriter` is interesting, but maybe not needed here - if we get a Pandas DataFrame type, it has a method called `to_markdown()` that makes it exceedingly simple to generate a markdown representation of the data

szaghikian-rasgo: Another option, if we decide that the output of `to_markdown()` is too loud, would be to directly call the library supporting the method: [tabulate](https://pypi.org/project/tabulate/), which is actually a great pretty data printing library that can be called directly on a Pandas DataFrame

szaghikian-rasgo:
@alexwolf22 check it out - changed the doc generator to do lots of the stuff we talked about. What u think?

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+        t1,# md.text(transform_yaml['preview-data']) if 'preview-data' in transform_yaml else None,
+        md.text(TRANSFORMED_DATA_PREVIEW_TEXT) if 'preview-data' in transform_yaml else None,
         md.python_code(transform_yaml['example_code']),
+        t2,# md.text(transform_yaml['post-transform-data']) if 'preview-data' in transform_yaml else None,
         md.h2('Source Code'),
         md.github_url(transform_type_dir_name, transform_name),
         '',
     ]
-    return '\n\n'.join(markdown_elements)
+    return '\n\n'.join(filter(lambda x: x!=None, markdown_elements))
```
alexwolf22, line 88: Why are we checking if on `None` here? Shouldn't every one of those values not be none?

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
 
+def _generate_dataset_markdowns(
+    rasgo: Any,
```
alexwolf22, line 92: We can set this type to `from pyrasgo.rasgo import Rasgo`

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+    dataset_id: int,
+    output_cols: List[str],
+    transform_details: Dict
+) -> List[str]:
+    """Creates markdown representation of both a pre-transform Rasgo Dataset
+    and a transformed Rasgo Dataset
+
+    Arguments:
+        :dataset_id: int: A Rasgo Community Dataset ID matching the columns named
```
alexwolf22, line 101: Could we remove the prefix `:` for params docs here? Makes it less legible

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+    output_cols: List[str],
+    transform_details: Dict
+) -> List[str]:
+    """Creates markdown representation of both a pre-transform Rasgo Dataset
+    and a transformed Rasgo Dataset
+
+    Arguments:
+        :dataset_id: int: A Rasgo Community Dataset ID matching the columns named 
+        in the `example_code` section of the YAML doc
+        :output_cols: List[str]: A list of column names to include in the output
```
alexwolf22, line 103: Isn't this the list of columns to filter from the input dataset? Not output? Var name and description is misleading

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+        YAML representation
+        :transform_args: dict: TODO what do we want here? How to build the transform from args
+    """
+    ds = rasgo.get.dataset(dataset_id)
+    ds_md = ds.preview()[output_cols].to_markdown()
```
alexwolf22, line 108: I want to talk more about use `df.to_markdown()` vs `MarkdownTableWriter` lib, vs [`tabulate`](https://pypi.org/project/pytablewriter/)`

If we think about the user goal here, it is to make the documentation about the transform as clear as possible.

Given the tables represent datasets, and not a table of args, it may look nicer to use `tabulate` strings over Markdown tables in the docs for datasets.

Also, another thing to consider, is that we want to minimize the amount of 3rd party libraries in our open source packages. Using `df.to_makedown()` would remove the requirement of using `MarkdownTableWriter` for making the transform args table current listed as a requirement in the repo too.

Would be good to sync up with @cpdough to see which one we should move forward with. It would be helpful to see to a comparison of doing the markdown table using tabulate would look like.

python/doc_generator.py
```diff
@@ -47,13 +75,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
 if __name__ == '__main__':
-    save_transform_docs()
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument('pyrasgo-api-key', action='store')
+    args = arg_parser.parse_args()
+
+
+    save_transform_docs(
+        rasgo_api_key=getattr(args, 'pyrasgo-api-key')
+    )
```
alexwolf22, line 126: Would be good to allow the user to specify which rasgo domain to hit (local vs staging vs prod).

Check out `python/publish_transforms.py` to see how we handle this. You could add the arg parse below and then use the function `utils.set_rasgo_domain_env(rasgo_domain)`

```python
arg_parser.add_argument('-d', '--rasgo-domain',
                        action='store',
                        default='production',
                        choices=['local', 'staging', 'production'],
                        help="Rasgo Environment to connect to. Sets env var "
                             "'RASGO_DOMAIN' for PyRasgo url dispatching "
                             "(local, staging, or production).")
```

python/doc_generator.py
```diff
@@ -47,13 +76,51 @@ def _get_transform_markdown(transform_yaml: Dict,
         md.h2('Parameters'),
         md.table(transform_yaml['arguments']),
         md.h2('Example'),
+        md.text(DATASET_PREVIEW_TEXT) if 'preview-data' in transform_yaml else None,
+        md.python_code(DATASET_PREVIEW_CODE) if 'preview-data' in transform_yaml else None,
+        t1,# md.text(transform_yaml['preview-data']) if 'preview-data' in transform_yaml else None,
+        md.text(TRANSFORMED_DATA_PREVIEW_TEXT) if 'preview-data' in transform_yaml else None,
         md.python_code(transform_yaml['example_code']),
```
alexwolf22, line 83: One thing about the example code...

I like it would be ideal to have a `1 to 1` relationship of the example code to create the example output df and the one shown.

Since we are generating the code and executing the transform to get the df, can't we just generate a nicely formatted string of python code as well?

This would also make writing the YAML easier, as someone won't need to supply `transform_yaml['example_code']` anymore.

python/doc_generator.py
```diff
@@ -1,18 +1,34 @@
...
+
+import pyrasgo
 
 from python import utils, markdown as md
 
 DOCS_DIR = utils.get_root_dir() / 'docs'
 
+DATASET_PREVIEW_TEXT = """Pull a source Dataset and preview it:"""
+
+DATASET_PREVIEW_CODE = """ds = rasgo.get.dataset(id)\nprint(ds.preview())"""
```
alexwolf22, line 13: As mentioned in another comment, I think it would be best if there is a `1 to 1` exact mapping of the code executed to create the output example df and the one shown in the docs, ideally using the same string to execute the python code with `eval(python_code)`.

What do you think about this?

python/doc_generator.py
```diff
@@ -32,13 +49,25 @@ def save_transform_docs() -> None:
...
 
 
-def _get_transform_markdown(transform_yaml: Dict,
-                            transform_type_dir_name: str,
-                            transform_name: str) -> str:
+def _get_transform_markdown(
+    transform_yaml: Dict,
+    transform_type_dir_name: str,
+    transform_name: str,
+    rasgo_connection: Any
```
alexwolf22, line 56: We can set this type to `from pyrasgo.rasgo import Rasgo`

python/doc_generator.py
```diff
@@ -47,13 +76,51 @@ def _get_transform_markdown(transform_yaml: Dict,
...
+    ds = rasgo.get.dataset(dataset_id)
+    ds_md = ds.preview()[output_cols].to_markdown()
+
+    ds1 = ds.transform(
+        transform_name=transform_details["transform-name"],
+        arguments=transform_details["transform-args"]
+    )
+    ds1_md = ds1.preview().to_markdown()
+
+    return[ds_md, ds1_md]
```
alexwolf22, line 116: Since we're returning two strings every time let's change the return type to `Tuple[str, str]` instead of a list of strings of any size.

Need to change `[ds_md, ds1_md]` to `ds_md, ds1_md` as well

rasgotransforms/rasgotransforms/column_transforms/math/math.yaml
```diff
@@ -18,4 +18,18 @@ example_code: |
...
-  ds2.preview()
\ No newline at end of file
+  ds2.preview()
+dataset-id: 74
+output-cols:
+  - SALESAMOUNT
+  - TAXAMT
+  - DUEDATE
+transform-details:
+  transform-name: math
```
alexwolf22, line 28: I don't think we need the transform name here, since we can get the value from the `file_name` when looping through the transforms while generating docs.

would be better to remove it so ppl don;lt need to type extra info for each transform YAML.

alexwolf22 (changes requested):
Great job getting this update out quick with the auto python df generator!

Added a bunch of comments want to talk about and potentially implement.

python/doc_generator.py
szaghikian-rasgo: Cool. `tabulate` looks way better in text, but way worse in markdown:
image
Tabulate on top here. Looks pretty attractive.

image
Tabulate on top here, too, but it's just gibberish, while the `to_markdown()` method looks pretty nice

python/doc_generator.py
szaghikian-rasgo: But it's also the columns seen in the output of the dataset preview, which is why I called it output_cols

python/doc_generator.py
szaghikian-rasgo: yeah this makes sense, and I actually had it in the first pass, but is anything in this code dependent on the domain? Community dataset IDs will remain the same, and we're not writing anything out to a place where we need the domain to determine the path?

python/doc_generator.py
szaghikian-rasgo: I wanted a list but I actually think you're right - structured and ordered would make more sense for this return data type than unstructured and ordered

PR #123, opened 2022-03-01
Sz/dr/a couple data samples

szaghikian-rasgo:
Guess what? The other PR is pretty complex and needs a day - 2 days of additional work. The good news - it'll be pretty cool when it's done. The bad news - I'm moving on to other stuff and we need to deliver some value. This is a PR that puts up some data samples in our documents in some of the places that could be useful for users to see them. This is a simple DR focused on seeing if users (or Patrick or Prof or Josh) like having the data samples around and think they're worth expanding.


griffatrasgo (approved):
Good one-day compromise. Nice killer instinct. Maybe shoot one of these doc pages over to Josh, Professor, Nick for feedback after this is merged.

szaghikian-rasgo:
Also, @cpdough, SQL file URLs are accurate in this repo, with the exception of datetrunc and label_encode. Those are incorrect because we split the SQL out by DB, and there is no default file - I added those defaults just now in this PR. That means when we push this branch to the RasgoDocs repo, we should have accurate URLs for 100% of transforms

griffatrasgo:
> Also, @cpdough, SQL file URLs are accurate in this repo, with the exception of datetrunc and label_encode. Those are incorrect because we split the SQL out by DB, and there is no default file - I added those defaults just now in this PR. That means when we push this branch to the RasgoDocs repo, we should have accurate URLs for 100% of transforms

@szaghikian-rasgo  Those files were intentionally missing because ansi sql has no support for date_trunc or array_agg. Adding those files back in is the wrong move - they will cause errors when you wire up postgres.

Something closer to the right move (up for debate), might be:
- rewrite those files to use some ansii compliant sql... good luck
- update our docs url code to act more like the serving code: check for a snowflake folder first, then default to the main folder only if a SF folder isn't found.
- leave those blank for this PR and deal with it later

I'd vote for 2 or 3, but I don't think this PR should ship with those files in it. It will create execution errors that everyone may notice to fix docs errors that few people will

griffatrasgo (changes requested):
Kill the datetrunc and labelencode files before merging. (See note)

alexwolf22 (approved):

griffatrasgo (approved):

PR #126, opened 2022-03-08
Make all transform docs into one dir

alexwolf22:
This PR is for removing the grouping of `table` `row`, and `column` transform types in organization for the docs.

We are doing this since transforms are now organized by tags.


cpdough:
In case you haven’t - can you make sure links to source code in docs are updated to match?

griffatrasgo (approved):
Looks good.

Per our kickoff: actual .sql/.yml structure will maintain the existing folder structure until we're ready to make a corresponding change to rasgql.

PR #131, opened 2022-03-15
allow no group by and add entropy calculations

RasgoDerek:
allows you to leave off group_by argument and allows you to get an ENTROPY calc for a column.  This change has two caveats:
1. The system can't stop you from being dumb and trying to get single column agg values and add a group by. If you do that you get invalid sql and it will fail on application to a dataset.
2. The template will break if a user tries to use ENTROPY in a DB environment that does not have `ratio_to_report` function

Test1 (works):

input:
```
---
aggregations:
  PRODUCTCATEGORYALTERNATEKEY:
  - SUM
  - ENTROPY
  PRODUCTCATEGORYKEY:
  - SUM
  - ENTROPY
source_table: ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY`

```

sql:
```
        with
    PRODUCTCATEGORYALTERNATEKEY_COUNTS as (
        select PRODUCTCATEGORYALTERNATEKEY, count(*) as c from ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY group by 1
    ),
    PRODUCTCATEGORYALTERNATEKEY_RATIOS as (
        select ratio_to_report(c) over() p_PRODUCTCATEGORYALTERNATEKEY from PRODUCTCATEGORYALTERNATEKEY_COUNTS
    ) ,
    PRODUCTCATEGORYKEY_COUNTS as (
        select PRODUCTCATEGORYKEY, count(*) as c from ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY group by 1
    ),
    PRODUCTCATEGORYKEY_RATIOS as (
        select ratio_to_report(c) over() p_PRODUCTCATEGORYKEY from PRODUCTCATEGORYKEY_COUNTS
    ) SELECT
SUM(PRODUCTCATEGORYALTERNATEKEY) as PRODUCTCATEGORYALTERNATEKEY_SUM,-sum(p_PRODUCTCATEGORYALTERNATEKEY*log(2,p_PRODUCTCATEGORYALTERNATEKEY)) as PRODUCTCATEGORYALTERNATEKEY_entropy,SUM(PRODUCTCATEGORYKEY) as PRODUCTCATEGORYKEY_SUM,-sum(p_PRODUCTCATEGORYKEY*log(2,p_PRODUCTCATEGORYKEY)) as PRODUCTCATEGORYKEY_entropy
FROM ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY, PRODUCTCATEGORYALTERNATEKEY_RATIOS, PRODUCTCATEGORYKEY_RATIOS
```

Output:
```
PRODUCTCATEGORYALTERNATEKEY_SUM	PRODUCTCATEGORYALTERNATEKEY_ENTROPY	PRODUCTCATEGORYKEY_SUM	PRODUCTCATEGORYKEY_ENTROPY
160	32	160	32
```

Test 2:
input:
```
---
group_by:
- ENGLISHPRODUCTCATEGORYNAME
aggregations:
  PRODUCTCATEGORYKEY:
  - SUM
  - AVG
  PRODUCTCATEGORYALTERNATEKEY:
  - SUM
  - AVG
source_table: ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY

```

sql:
```
SELECT
ENGLISHPRODUCTCATEGORYNAME,SUM(PRODUCTCATEGORYALTERNATEKEY) as PRODUCTCATEGORYALTERNATEKEY_SUM,AVG(PRODUCTCATEGORYALTERNATEKEY) as PRODUCTCATEGORYALTERNATEKEY_AVG,SUM(PRODUCTCATEGORYKEY) as PRODUCTCATEGORYKEY_SUM,AVG(PRODUCTCATEGORYKEY) as PRODUCTCATEGORYKEY_AVG
FROM ADVENTUREWORKS.PUBLIC.DIMPRODUCTCATEGORY
GROUP BY ENGLISHPRODUCTCATEGORYNAME
```

output:
```
ENGLISHPRODUCTCATEGORYNAME	PRODUCTCATEGORYALTERNATEKEY_SUM	PRODUCTCATEGORYALTERNATEKEY_AVG	PRODUCTCATEGORYKEY_SUM	PRODUCTCATEGORYKEY_AVG
Bikes	1	1.000000	1	1.000000
Components	2	2.000000	2	2.000000
Clothing	3	3.000000	3	3.000000
Accessories	4	4.000000	4	4.000000
```

### DIFF
--- a/docs/aggregate.md
+++ b/docs/aggregate.md
@@ -8,7 +8,7 @@ Groups rows by the group_by items applying aggregations functions for the result
 
 |   Argument   |    Type     |                                                             Description                                                             | Is Optional |
 | ------------ | ----------- | ----------------------------------------------------------------------------------------------------------------------------------- | ----------- |
-| group_by     | column_list | Columns to group by                                                                                                                 |             |
+| group_by     | column_list | Columns to group by                                                                                                                 | True        |
 | aggregations | agg_dict    | Aggregations to apply for other columns. Dict keys are column names, and values are a list of aggegations to apply for that column. |             |
 
 
@@ -22,6 +22,10 @@ ds2 = ds.aggregate(group_by=['FIPS'], aggregations={
           'COL_2': ['SUM', 'AVG']
       })
 ds2.preview()
+
+ds3 = ds.aggregate(aggregations={'COL_1': ['SUM']})
+
+ds4 = ds.aggregate(aggregations={'DISTINCT COL_1': ['SUM']})
 ```
 
 ## Source Code
--- a/docs/apply.md
+++ b/docs/apply.md
@@ -2,7 +2,7 @@
 
 # apply
 
-A transform that accepts a custom template to execute. Must use the sql template argument `source_table` to reference the Rasgo dataset which will serve as the base of any SELECT
+A transform that accepts a custom template to execute. Must use the sql template argument `source_table` to reference the Rasgo dataset which will serve as the base of any SELECT. Other templatized parameters must also be passed in as kwargs.
 
 ## Parameters
 
@@ -17,9 +17,11 @@ A transform that accepts a custom template to execute. Must use the sql template
 ds = rasgo.get.dataset(id)
 
 ds2 = ds.apply(
-  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17'
+  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17',
+  source_table='db.schema.table'
 )
 ds2.preview()
+#TODO Add args 
 ```
 
 ## Source Code
--- a/docs/apply.md
+++ b/docs/apply.md
@@ -17,11 +17,16 @@ A transform that accepts a custom template to execute. Must use the sql template
 ds = rasgo.get.dataset(id)
 
 ds2 = ds.apply(
-  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17',
-  source_table='db.schema.table'
+  sql='SELECT * FROM {{ source_table }} WHERE {{ where_column }} = {{ where_value }}',
+  where_column='NEW_LEADS_COL',
+  where_value=100
+)
+
+ds2 = ds.apply(
+  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17'
 )
 ds2.preview()
-#TODO Add args 
+
 ```
 
 ## Source Code
--- a/rasgotransforms/rasgotransforms/table_transforms/aggregate/aggregate.sql
+++ b/rasgotransforms/rasgotransforms/table_transforms/aggregate/aggregate.sql
@@ -1,13 +1,40 @@
+{%- set entropy_aggs = {} -%}
+{%- set from_tables = [source_table] -%}
+{%- for col, aggs in aggregations.items() -%}
+    {%- if 'ENTROPY' in aggs -%}
+        {%- set _ = entropy_aggs.update({col: aggs}) -%}
+    {%- endif -%}
+{%- endfor -%}
+
+{%- for col, aggs in entropy_aggs.items() -%}
+    {%- if loop.first %}
+        with
+    {%- endif %}
+    {{ col }}_COUNTS as (
+        select {{col}}, count(*) as c from {{ source_table }} group by 1
+    ),
+    {{ col }}_RATIOS as (
+        select ratio_to_report(c) over() p_{{col}} from {{ col }}_COUNTS
+    ) {{ ',' if not loop.last }}
+    {%- set _ = from_tables.append(col+'_RATIOS')  -%}
+{%- endfor -%}
 SELECT
-{%- for group_item in group_by %}
+{% for group_item in group_by -%}
     {{ group_item }},
 {%- endfor -%}
 
-{%- for col, aggs in aggregations.items() %}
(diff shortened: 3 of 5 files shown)


docs/apply.md
```diff
@@ -17,9 +17,11 @@ A transform that accepts a custom template to execute. Must use the sql template
 ds = rasgo.get.dataset(id)
 
 ds2 = ds.apply(
-  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17'
+  sql='SELECT * FROM {{ source_table }} WHERE COLUMNVALUE = I17',
+  source_table='db.schema.table'
 )
```
alexwolf22, line 22: So in `PyRasgo` the `source_table` here gets out-populated into the args dict like below, so it doesn't make sense to add that as a kwarg here for example code.

```python
args['source_table'] = ds.fqtn
```

I think it will be better to instead have an Apply transform with two/three jinja args. Mabe something like below
```python
ds2 = ds.apply(
  sql='SELECT * FROM {{ source_table }} WHERE {{ where_column }} = {{ where_value }}',
  where_column='NEW_LEADS_COL',
  where_value=100
)
```

docs/apply.md
```diff
@@ -17,9 +17,11 @@ A transform that accepts a custom template to execute. Must use the sql template
...
 ds2.preview()
+#TODO Add args
```
alexwolf22, line 24: did you mean to keep this here?

alexwolf22 (changes requested):

docs/apply.md
RasgoDerek: Good feedback thanks

docs/apply.md
RasgoDerek: 👍

rasgotransforms/rasgotransforms/table_transforms/aggregate/aggregate.sql
```diff
@@ -1,13 +1,40 @@
...
+        with
+    {%- endif %}
+    {{ col }}_COUNTS as (
+        select {{col}}, count(*) as c from {{ source_table }} group by 1
+    ),
+    {{ col }}_RATIOS as (
+        select ratio_to_report(c) over() p_{{col}} from {{ col }}_COUNTS
+    ) {{ ',' if not loop.last }}
+    {%- set _ = from_tables.append(col+'_RATIOS')  -%}
+{%- endfor -%}
```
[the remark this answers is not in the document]
RasgoDerek: Nested CTEs are legal

docs/aggregate.md
```diff
@@ -8,7 +8,7 @@ Groups rows by the group_by items applying aggregations functions for the result
 
 |   Argument   |    Type     |                                                             Description                                                             | Is Optional |
 | ------------ | ----------- | ----------------------------------------------------------------------------------------------------------------------------------- | ----------- |
-| group_by     | column_list | Columns to group by                                                                                                                 |             |
```
RasgoDerek, line 11: @alexwolf22 I didn't make these changes but they appear to be different as a result of some automated generations between my view and main.  Can you offer some insights about what this should look like here?

docs/aggregate.md
```diff
@@ -22,6 +22,10 @@ ds2 = ds.aggregate(group_by=['FIPS'], aggregations={
           'COL_2': ['SUM', 'AVG']
       })
 ds2.preview()
+
+ds3 = ds.aggregate(aggregations={'COL_1': ['SUM']})
+
+ds4 = ds.aggregate(aggregations={'DISTINCT COL_1': ['SUM']})
```
cpdough, line 28: i think inputting DISTINCT as part of the column specification might be confusing, and also break the UI input, vs treating it as a new agg method

cpdough:
> The template will break if a user tries to use ENTROPY in a DB environment that does not have ratio_to_report function
For this reason we should make this version snowflake specific and leave entropy out of the general aggregate function template

docs/aggregate.md
RasgoDerek: you can do  `select sum(distinct column)` and same with all the other aggregations.  If we relegate it to the aggregation functions, I think that means we have to disallow that use case. Another option is to make the aggregation object a little more complicated and do something like 

```
aggregations={
  'Col1' : {
    aggregations: ['SUM', 'AVG', ...],
    distinct: true
    }
}
```

But that would seem to imply that you want the distinct aggregation of all the aggregations you select.  Is that preferable?  It would also make it legal to add `Col1` twice, once as a distinct set of aggregations and another as not. an if you just want to `SELECT DISTINCT Col1 FROM wherever` you could just pass in `distinct: true` with no aggregations.

But that's getting way more complicated than usable.  Thoughts?

RasgoDerek:
>we should make this version snowflake specific and leave entropy out of the general aggregate function template

@cpdough I don't see those distinctions being made in the transforms here, I thought that was a rasgoQL specific implementation

docs/aggregate.md
alexwolf22: Yeah since you set `group_by` transform param to True in the YAML, it added that to the Markdown.

Markdown file seems right here.

RasgoDerek:
Another option is to implement a feature that defines functions in whatever db they may need to exist in, or just implement whatever query underlies `ratio_to_report` directly instead of relying on snowflake magic.

Of note, this will work in bigquery as written.

PR #133, opened 2022-03-16
Jberry timeseriesagg

itsamejoshab:
- changed methodology to have only 1 subquery per-offset rather than per combination of column
- slight renaming of output variables for convenience


cpdough (approved):

itsamejoshab:
performance improvements to timeseries_agg

PR #138, opened 2022-03-21
add alias support for datediff on all dbs

cpdough:


griffatrasgo (approved):
Nice change. Taught me new jinja syntax in the process

PR #148, opened 2022-04-04
Remove transform_type 

griffatrasgo:
We are deprecating the concept of "transform types" and moving to tagging of transforms. Docs and yaml files have already been updates to reflect this. This PR makes the structural changes to remove the concept of transform_type from transforms.

This PR:
- moves all template files into a single "transforms" folder and deletes the outdated transform type folders
- removes the `transform_type` property from `TransformTemplate` class
- updates all scripts that serve or publish docs to reference the new structure

BONUS
- reads a single yaml file per transforms (in the main folder) - so transforms with multiple versions per DW now only need separate .sql files, but share the same .yaml file


alexwolf22 (approved):
Looks good nice job with this PR!

I'm sure you've done so already, but I would double-check to make sure all the functionality for publishing, generating docs, and sever transforms for RQL works as expected.

python/doc_generator.py
```diff
@@ -20,35 +20,32 @@ def save_transform_docs() -> None:
...
+            transform_dir_name="transforms",
+            transform_name=transform_name,
+            dw_type=dw_type
+        )
+        # Write Transform Markdown to Docs directory
+        md_file_path = DOCS_DIR / f"{transform_name}.md"
+        os.makedirs(md_file_path.parent, exist_ok=True)
+        with md_file_path.open('w') as fp:
+            print(f"Writing Markdown docs at {md_file_path}\n")
+            fp.write(markdown)
```
ChrisGriffithRASGO, line 44: Just a nice to know, this type of stuff can be done directly with the `Path` objects now. 

```suggestion
        md_file_path.parent.mkdir(exist_ok=True) 
        print(f"Writing Markdown docs at {md_file_path}\n")
        md_file_path.write_text(markdown)
```

griffatrasgo:
> Looks good nice job with this PR!
> 
> I'm sure you've done so already, but I would double-check to make sure all the functionality for publishing, generating docs, and sever transforms for RQL works as expected.

I was thinking I'd save a few minutes by skipping this part and just deploying blindly.

PR #150, opened 2022-04-06
Remove first column of histogram and bar chart

phillip-rasgo:
Removes the first column in the "histogram" and "bar_chart" select statements, so that the first two columns define teh X Axis bucket boundaries.  I have no idea if this SQL will execute successfully.


cnash6 (approved):
if you say it works let's test it!

PR #151, opened 2022-04-07
Cool stuff

cpdough:
re-arranging the templates to match the right user description of what that template should do


cpdough:
note: line chart still needs to be expanded to support multiple columns. this PR does not do that.

rasgotransforms/rasgotransforms/transforms/line_chart/line_chart.sql
```diff
@@ -1,11 +1,38 @@
...
+  ,MIN_VAL
+  ,MAX_VAL
+  ,BUCKET_SIZE
+FROM
+  COUNTS
+CROSS JOIN EDGES)
+SELECT
+  MIN_VAL+((HIST_BUCKET-1)*BUCKET_SIZE) AS {{ column }}_MIN
+  ,MIN_VAL+(HIST_BUCKET*BUCKET_SIZE) AS {{ column }}_MAX
+  ,SUM(REC_CT) AS RECORD_COUNT
```
phillip-rasgo, line 34: Should this be the mean or median of column values in the bucket, since this template is supposed to generate an approximation of the series?

rasgotransforms/rasgotransforms/transforms/line_chart/line_chart.sql
```diff
@@ -1,11 +1,38 @@
...
+  ,REC_CT
+  ,WIDTH_BUCKET(VAL, MIN_VAL, MAX_VAL, {{ bucket_count }}) AS HIST_BUCKET
+  ,MIN_VAL
+  ,MAX_VAL
+  ,BUCKET_SIZE
+FROM
+  COUNTS
+CROSS JOIN EDGES)
+SELECT
+  MIN_VAL+((HIST_BUCKET-1)*BUCKET_SIZE) AS {{ column }}_MIN
```
phillip-rasgo, line 32: Should this bucket across a separate dimension column, to use as the X axis?

phillip-rasgo (commented):
histogram and bar_chart look useful.  line_chart looks like it's running the same logic as histogram though, rather than approximating a column's values over a second (dimension) column.

PR #152, opened 2022-04-07
RAS-2783 Fix list_transforms for postgresql

ChrisGriffithRASGO:
The root issue of this is when calling `list_transforms` and there is no base `sql` file or dialect specific override for a transform.

```
import rasgoql
creds = rasgoql.PostgresCredentials("1", "1", "1", "1", "1", "1")
rql = rasgoql.connect(creds)
rql.list_transforms()
# Traceback (most recent call last): ...
# FileNotFoundError 
```

The fix is to specifically ignore FileNotFoundError as they are expected in this case.


griffatrasgo (approved):
Nice solve

PR #155, opened 2022-04-11
Add support for Redshift

ChrisGriffithRASGO:
These are a copy of the files from postgresql, as redshift has the same syntax. Confirmed `dateadd` working directly, and that the docs are the same for [`date_part` ](https://docs.aws.amazon.com/redshift/latest/dg/r_DATE_PART_function.html) and [`date_trunc`](https://docs.aws.amazon.com/redshift/latest/dg/r_DATE_TRUNC.html)

### DIFF
--- a/rasgotransforms/rasgotransforms/main.py
+++ b/rasgotransforms/rasgotransforms/main.py
@@ -23,7 +23,7 @@ class Datawarehouse(Enum):
     SNOWFLAKE = "snowflake"
     POSTGRESQL = "postgresql"
     MYSQL = "mysql"
-
+    REDSHIFT = "redshift"
 
 class TransformTemplate:
     """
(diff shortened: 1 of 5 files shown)


rasgotransforms/rasgotransforms/main.py
```diff
@@ -148,7 +148,7 @@ def _load_all_yaml_files(datawarehouse: str) -> Dict[str, Dict]:
         try:
             transform_data = _read_yaml(transform_yaml_path)
             transform_yamls[transform_name] = transform_data
-        except Exception:
+        except FileNotFoundError:
```
ChrisGriffithRASGO, line 151: This is technically out of scope  for this PR, can be removed if wanted. 

This is simply a safety measure to limit catching the expected error in this case so we have insight if something else goes wrong.

griffatrasgo (approved):

PR #157, opened 2022-04-12
make a 2-d density heatmap

cpdough:
builds a 2-d density heatmap in SQL. screenshot to see resultant data

Screen Shot 2022-04-12 at 10 03 27 AM


szaghikian-rasgo (approved):

phillip-rasgo (approved):
200 buckets per axis would mean 40,000 rows of data.  I think that's going to be too many...

cpdough:
changed to 100

PR #162, opened 2022-04-14
Adding pre-commit and black code formatter

ChrisGriffithRASGO:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+      exclude: 'rasgotransforms/rasgotransforms/transforms'
+    - id: check-executables-have-shebangs
+    - id: end-of-file-fixer
+      exclude: 'rasgotransforms/rasgotransforms/transforms'
+    - id: no-commit-to-branch
+  - repo: https://github.com/psf/black
+    rev: 22.3.0
+    hooks:
+      - id: black
+        args: [--line-length=120, --skip-string-normalization]
```
alexwolf22, line 33: Any reason to use these and not the defaults? Never used black in detail before.

Are there other options we may want to change?

README.md
```diff
@@ -17,11 +17,11 @@ Rasgo Transforms provide jinja SQL templates that can be applied to your data us
 # Running Transforms
 
 Rasgo Transforms can be applied via:
-- your Rasgo Feature Store account ([pyrasgo](https://pypi.org/project/pyrasgo/) - Account required) 
+- your Rasgo Feature Store account ([pyrasgo](https://pypi.org/project/pyrasgo/) - Account required)
```
alexwolf22, line 20: @cnash6 We're moving away from calling Rasgo a Feature store right? Shouldn't we change the text here then?

python/constants.py
```diff
@@ -16,11 +17,7 @@ class PyRasgoEnvironment(Enum):
 # Different Types of Transforms
 # This var is used to find all transforms of a type in the
 # directory <root>/<transform_type>_transforms/...
-TRANSFORM_TYPES = [
-    'column',
-    'row',
-    'table'
-]
+TRANSFORM_TYPES = ['column', 'row', 'table']
```
alexwolf22, line 20: @ChrisGriffithRASGO We used to use transform types for file directory structure but recently removed them and replaced it with transform 'tags'

We can delete this part of code if you don't see it used anywhere.

python/utils.py
```diff
@@ -214,9 +213,11 @@ def _transform_args_have_changed(
...
         # if so, transform needs versioning
         db_arg = db_args_w_name[0]
-        if db_arg.description != yaml_arg['description'] or \
-                db_arg.is_optional != yaml_arg.get('is_optional', False) or \
-                db_arg.type != yaml_arg['type']:
+        if (
+            db_arg.description != yaml_arg['description']
+            or db_arg.is_optional != yaml_arg.get('is_optional', False)
+            or db_arg.type != yaml_arg['type']
+        ):
```
alexwolf22, line 220: Wow never knew you could organize long `if` statements like this.

Like that way more than using the slashes

griffatrasgo (approved):

alexwolf22 (commented):
So I don't have enough experience with the pre-commit stuff to know if this is what we want for the package exactly.

Overall I like the black code formatter; there are definitely a few times when it formatted code how I didn't like it, but would be good to get consistency across the codebase, especially for public repo.

Made a few other comments as well.

alexwolf22:
Also what was the reason again for using pre-commit over Github actions?

We have one GitHub action for auto-generating the transform docs/markdown; may be worth working think about adding that as a pre-commit instead. If we do this people would catch errors more easily if they had committed bad YAML. Thoughts @cnash6

ChrisGriffithRASGO:
> Also what was the reason again for using pre-commit over Github actions?
> 
> We have one GitHub action for auto-generating the transform docs/markdown; may be worth working think about adding that as a pre-commit instead. If we do this people would catch errors more easily if they had committed bad YAML. Thoughts @cnash6

pre-commit.ci is free for open source and quick drop in so no work involved. The github actions in the marketplace for pre-commit even have self depreciated in favor of of pre-commit.ci https://github.com/pre-commit/action

If someone submits bad YAML in this case the `check-yaml` pre-commit hook would catch it and report as an unsuccessful check. Which you can see when hitting `Show all checks` and for example would link you too to associated run. Example RasgoQL run https://results.pre-commit.ci/run/github/450978755/1649907419.NhiRXLnfTO2-e1Ofmi2zVQ

pre-commit.ci is created and run by a trusted python community member.  https://github.com/sponsors/asottile  I cannot vouce for security practices using it on a private repo but like the idea of using it for anything open source.

.pre-commit-config.yaml
ChrisGriffithRASGO: Line length devs can fight over all day long, going through our code bases, it seemed that most things were soft wrapped at 120 characters so set it there (Both PyCharm and VSCode have ruler marks at 120, so it's a common setting.) 

`--skip-string-normalization` we will hopefully remove when we have time (just means all strings will go to `"` instead of `'` and usually want to do some manual cleanup after it.)

python/constants.py
ChrisGriffithRASGO: Avoiding logic changes with this PR. Added as issue in linter. Good insight!

PR #165, opened 2022-04-18
what if the distinct values in a column are numeric?

szaghikian-rasgo:
Creating this graph's data entails unioning distinct values from multiple columns into either a `sourced_node` or a `dest_node`. Obviously, if those distinct values are both numeric and varchar types and we don't cast them, we'll get a failure in the template. This PR resolves that issue.

### DIFF
--- a/rasgotransforms/rasgotransforms/transforms/funnel/funnel.sql
+++ b/rasgotransforms/rasgotransforms/transforms/funnel/funnel.sql
@@ -1,7 +1,7 @@
-{%- for i in range((hierarchy|length) - 1) -%}
+{%- for i in range((stage|length) - 1) -%}
     SELECT
-    CAST({{ hierarchy[i] }} AS STRING) AS SOURCE_NODE,
-    CAST({{ hierarchy[i+1] }} AS STRING) AS DEST_NODE,
+    CAST({{ stage[i] }} AS STRING) AS SOURCE_NODE,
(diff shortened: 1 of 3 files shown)


rasgotransforms/rasgotransforms/transforms/funnel/funnel.sql
```diff
@@ -1,7 +1,7 @@
 {%- for i in range((hierarchy|length) - 1) -%}
     SELECT
-    {{ hierarchy[i] }} AS SOURCE_NODE,
-    {{ hierarchy[i+1] }} AS DEST_NODE,
+    CAST({{ hierarchy[i] }} AS STRING) AS SOURCE_NODE,
```
griffatrasgo, line 3: Do we need to 0 pad the strings so we don't get sorting = 1, 11, 12, ..., 2?

Or will the UI handle that for us?

szaghikian-rasgo: the order of `source_node` and `dest_node` doesn't matter - the vis code in the UI matches sources and destinations

szaghikian-rasgo:
Questions for this PR @cpdough :
1. Do we want to change the name of the `heirarchy` argument?
2. The chart type is `sankey` and the transform name is `funnel`. Do we want to invert or change these?

PR #167, opened 2022-04-19
let's open up the option of dates

szaghikian-rasgo:
Currently, our `line_chart` template will break if someone wants to use a date as their x-axis. That's not super lit, because one of the primary thing's you might want to do with a chart like this is chart a pattern over time. This change inspects the information schema to determine if we should treat the axis as a `DATE` or `TIMESTAMP` or something else - then we treat it that way.


rasgotransforms/rasgotransforms/transforms/line_chart/line_chart.yaml
```diff
@@ -8,6 +8,9 @@ arguments:
   axis:
     type: column
     description: axis column
+  axis_is_date:
```
griffatrasgo, line 11: Is there a way to make this optional and default to FALSE?

szaghikian-rasgo: >I'm going to make this pencil disappear

PR #169, opened 2022-04-20
added replace_string and regexp_replace

kimballh:
I added two transform templates, one for REGEXP_REPLACE (regexp_replace) and one for REPLACE (replace_string)


cpdough:
Nice work @kimballh - question is, do we need both of these? Is there user functionality in replace that you can’t get in regex replace? Or, is there a way to consolidate the two templates into one and pick the right sql function depending on the inputs?

cpdough:
Nice work @kimballh - question is, do we need both of these? Is there user functionality in replace that you can’t get in regex replace? Or, is there a way to consolidate the two templates into one and pick the right sql function depending on the inputs?

cpdough:
i love option 2 to give the flexibility to the user for when they need it, but keep it to a single transform in our library

PR #170, opened 2022-04-20
enhanced sample to include filter statements

kimballh:
Added filter_statements as an optional parameter for the .sample transformation

### DIFF
--- a/docs/sample.md
+++ b/docs/sample.md
@@ -6,9 +6,10 @@ Take a sample of a dataset using a specific number of rows or a probability that
 
 ## Parameters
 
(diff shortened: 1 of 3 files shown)


cpdough:
Hey kimball, would this apply the filter before or after sampling? I’m thinking it needs to be before in order to make sure that the sample is on the subsetted data

kimballh:
@cpdough I tested it with the where clause after and still go the same number of rows, so I assumed it was happening in order, but after some research, I think that it won't calculate probabilities the right way when passing a float, so I changed it to do a sub-query before the sample now if filter_statements is passed in.

PR #171, opened 2022-04-21
added outliers transform

kimballh:
Added the outliers transform which supports IQR, Z-score, and manual thresholding methods for calculating outliers.

- For IQR I kept it simple (1.5 * IQR). Should we make the multiplier or the quartiles dynamic?
- For Z-score the user can provide the max Z-score (default is 2)
- For manual threshold, the user simply provides a min and a max
- All methods create a column OUTLIER_\<column\> which is a boolean flag for outliers
- IQR and Z-score methods make some extra columns, should we remove these?
    - IQR makes Q1_\<column\>, MEDIAN_\<column\>, and Q3_\<column\>
    - Z-score makes MEAN_\<column\>, STDDEV_\<column\>, and ZSCORE_\<column\>


kimballh:
Another thought, should I add a 'drop' parameter which drops the outliers, rather than just flag them in the OUTLIER column?

rasgotransforms/rasgotransforms/transforms/outliers/outliers.yaml
```diff
@@ -0,0 +1,56 @@
...
+    type: value
+    description: The maximum threshold for values that won't be flagged as outliers.
+      Required if method is "threshold".
+    is_optional: true
+  max_zscore:
+    type: value
+    description: The maximum Z-score for values which will not be flagged as an outlier. 
+      Default Value = 2
+    is_optional: true
+example_code: |
```
cpdough, line 47: @kimballh  i think the example_code needs to be updated to reflect the new arguments you added

kimballh: @cpdough Done, I added an example for all three methods.

PR #182, opened 2022-05-02
Add a funnel transform template

szaghikian-rasgo:
Addresses data needed for the funnel chart type. Currently, only accepts numeric data types where we can simply sum the values to get the total incidence count. 

Curious on usability @cpdough, do we want to allow this transform template to be used on numeric and categorical data? If so, I'll add a column type query and logic branch a la other transforms in this repo. 

Curious on data return structure from @phillip-rasgo, would you rather see funnel stage values column-wise, like this:
image

Or row-wise, like this:
image


rasgotransforms/rasgotransforms/transforms/funnel/funnel.sql
```diff
@@ -0,0 +1,15 @@
...
+FROM {{ source_table }}
+{{ "UNION ALL" if not loop.last else "" }}
+{% endfor %}
+
+-- OR, for a multi-column stage return:
+-- SELECT
+-- {%- for col_name in stage_columns -%}
+-- {{ "," if not loop.first else "" }}SUM({{ col_name }})
+-- {% endfor %}
+-- FROM {{ source_table }}
```
szaghikian-rasgo, line 15: Two options, for row-wise v column-wise returns.

rasgotransforms/rasgotransforms/transforms/funnel/funnel.yaml
```diff
@@ -0,0 +1,18 @@
...
+context:
+  # TODO: do we want to leave this as a sankey? No, right, it's a new chart type?
+  chart_type: funnel
+tags: 
+# TODO: this description is trash
+description: If you have numeric columns representing the number of times a particular outcome occurs in your data, this transform creates a ready-to-visualize dataset of the survival rate in the funnel
+arguments:
+  stage_columns:
+    type: column_list
+    # TOOD: this description is trash, too
```
szaghikian-rasgo, line 12: I'm cleaning up both of the descriptions here for clarity in the next commit

rasgotransforms/rasgotransforms/transforms/sankey/sankey.yaml
```diff
@@ -11,5 +11,5 @@ arguments:
 example_code: |
   ds = rasgo.get.dataset(id)
 
-  ds2 = ds.funnel(stage=["ENGLISHCOUNTRYREGIONNAME", "STATEPROVINCENAME", "CITY"])
+  ds2 = ds.sankey(stage=["ENGLISHCOUNTRYREGIONNAME", "STATEPROVINCENAME", "CITY"])
```
szaghikian-rasgo, line 14: This was an error in the sankey docs

PR #185, opened 2022-05-04
RAS-2641 Support Overwrite Columns Parameter

kimballh:
Added overwrite_column argument to:
- concat
- dateadd
- min max scaler
- rank
- standard scaler
- to date

The [ticket](https://linear.app/rasgoml/issue/RAS-2641/support-overwrite-columns=true-or-false-parameter-in-many-transforms) asked for datediff as well, but I didn't include it since the output is the combination of two date columns and is a different type. @cpdough Do you want me to add it as an option and just delete the columns if overwrite_columns=True?

### DIFF
--- a/docs/concat.md
+++ b/docs/concat.md
@@ -9,10 +9,11 @@ Pass in a list named "concat_list", containing the names of the columns and the 
 
 ## Parameters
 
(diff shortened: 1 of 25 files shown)


kimballh:
I also left out `impute` because overwriting the columns is already the default behavior unless `flag_missing_vals=True`

PR #202, opened 2022-05-17
change filter and sort args

szaghikian-rasgo:
Adding new arg types for ordering `ASC` or `DESC` - we now apparently have an arg type `sort` that will result in a dropdown from which people can select `ASC` or `DESC`. 

Also adding new filter functionality to all existing instances of `filter_statements` or similar args.

Also also deleting the bar chart and line chart transforms since we now have plot to replace them


docs/histogram.md
```diff
@@ -6,11 +6,11 @@ Analyze the value distribution of a single continuous variable by binning it and
...
 
-|       Name        |    Type     |                                  Description                                   | Is Optional |
-| ----------------- | ----------- | ------------------------------------------------------------------------------ | ----------- |
-| column            | column      | numeric column to use to generate the histogram                                |             |
-| filter_statements | string_list | List of SQL where statements to filter the table by, i.e. 'COLUMN IS NOT NULL' | True        |
-| num_buckets       | value       | max number of buckets to create; defaults to 200                               | True        |
+|       Name        |    Type     |                   Description                    | Is Optional |
+| ----------------- | ----------- | ------------------------------------------------ | ----------- |
+| column            | column      | numeric column to use to generate the histogram  |             |
+| filter_statements | filter_list | list of dictionaries representing filters        | True        |
```
cpdough, line 12: can we name this input 'filters' everywhere instead of filter_statements?

docs/plot.md
```diff
@@ -11,8 +11,8 @@ Visualize a dataset flexibly, depending on axes and metrics chosen
...
 | num_buckets       | value           | max number of buckets to create; defaults to 200                                                                                                                             | True        |
-| filter_statements | string_list     | List of SQL where statements to filter the table by, i.e. 'COLUMN IS NOT NULL'                                                                                               | True        |
-| order_direction   | string          | Either ASC or DESC, depending on if you'd like to order your bar chart X-axis returned in ascending or descending order                                                      | True        |
+| filter_statements | filter_list     | list of dictionaries representing filters                                                                                                                                    | True        |
+| order_direction   | sort            | Either ASC or DESC, depending on if you'd like to order your bar chart X-axis returned in ascending or descending order                                                      | True        |
```
cpdough, line 15: can we use the name to make it clear that this is an order_by for the x_axis? like 'x_axis_order' ?

rasgotransforms/rasgotransforms/transforms/histogram/histogram.sql
```diff
@@ -11,9 +11,19 @@ SELECT
   ,COUNT(1) AS REC_CT
 FROM
   (SELECT {{ column }}::float AS COL FROM {{ source_table }}
-  {%- if filter_statements is iterable -%}
-  {%- for filter_statement in filter_statements %}
-  {{ 'WHERE' if loop.first else 'AND' }} {{ filter_statement }}
+  {%- if filter_statements is defined and filter_statements %}
```
cpdough, line 14: is this a double existence check?

docs/histogram.md
cpdough: also sorry for the nitpick but let's change the description everywhere too since it will show up on the hover tooltip in UI:
'Filter logic on one or more columns. Can choose between a simple comparison filter or advanced filter using free text.'

cpdough (changes requested):

rasgotransforms/rasgotransforms/transforms/histogram/histogram.sql
szaghikian-rasgo: Not a double existence - `is defined` is a presence check, and `filter_statements` is a check to see if the iterable contains anything

rasgotransforms/rasgotransforms/transforms/histogram/histogram.sql
```diff
@@ -11,9 +11,19 @@ SELECT
...
+  {{ 'WHERE ' if oloop.first else ' AND ' }}
+  {%- if filter_block is not mapping -%}
+  {{ filter_block }}
+  {%- else -%}
+      {%- if filter_block['operator'] == 'CONTAINS' -%}
+  {{ filter_block['operator'] }}({{ filter_block['columnName'] }}, {{ filter_block['comparisonValue'] }})
+      {%- else -%}
+  {{ filter_block['columnName'] }} {{ filter_block['operator'] }} {{ filter_block['comparisonValue'] }}
+      {%- endif -%}
+  {%- endif -%}
```
ChrisGriffithRASGO, line 26:
```suggestion
  {%- if filters is defined and filters %}
      {% for filter_block in filters %}
          {%- set oloop = loop -%}
          {{ 'WHERE ' if oloop.first else ' AND ' }}
          {%- if filter_block is not mapping -%}
              {{ filter_block }}
          {%- else -%}
              {%- if filter_block['operator'] == 'CONTAINS' -%}
                  {{ filter_block['operator'] }}({{ filter_block['columnName'] }}, {{ filter_block['comparisonValue'] }})
               {%- else -%}
                  {{ filter_block['columnName'] }} {{ filter_block['operator'] }} {{ filter_block['comparisonValue'] }}
               {%- endif -%}
  {%- endif -%}
```

Would be nice to update these so the indentation matches the logic around it

szaghikian-rasgo: @ChrisGriffithRASGO check it now?

ChrisGriffithRASGO (approved):
Thank you for the pretties

PR #203, opened 2022-05-17
Utilize the `sort` arg

szaghikian-rasgo:
This PR should catch all places where we do an `order by` based on a transform argument. Let's revisit this when the argument type has been created.


szaghikian-rasgo:
Killing this PR in favor of possible future client-side sorting to allow for sorting by metric values or dimension values

PR #206, opened 2022-05-17
plot whitespace on categorical axis

szaghikian-rasgo:


szaghikian-rasgo:
Looked at the other transforms that had filter changed today. None have trailing whitespace issues

PR #208, opened 2022-05-18
Updating publish transforms to work with multiple dw_types

ChrisGriffithRASGO:
Adding support to publish transforms for each of the current supported DWs in RasgoAPI, as well as a generic one. Dependent on RasgoAPI and RasgoSDK changes.

(Seems like there are some pre-commit things that are not being run for this repo, so there is also some cleanup there. Need to figure out why pre-commit.ci isn't working)

### DIFF
--- a/python/constants.py
+++ b/python/constants.py
@@ -21,3 +21,5 @@ GITHUB_REPO_URL = "https://github.com/rasgointelligence/RasgoTransforms/blob/mai
 # TODO: When we support multiple DWs in the API
 # we'll need to refactor the functions that consume this
 RASGO_DATAWAREHOUSE = 'snowflake'
+
+COMMUNITY_ORGANIZATION_ID: int = 1
--- a/python/publish_transforms.py
+++ b/python/publish_transforms.py
@@ -60,21 +60,25 @@ def publish_transforms(rasgo_api_key: str, rasgo_domain: str) -> None:
                 continue
             if (name, dw) not in rasgo_transforms:
                 print(
-                    f"No transform with name '{name}' for {dw.capitalize()} warehouse found in Rasgo. "
+                    f"No transform with name '{name}' for {dw} warehouse found in Rasgo. "
                     f"Creating new '{transform_name}' transform "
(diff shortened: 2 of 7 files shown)


python/publish_transforms.py
```diff
@@ -30,7 +30,12 @@ def publish_transforms(rasgo_api_key: str, rasgo_domain: str) -> None:
 
     # Get all Rasgo Transforms
     rasgo = pyrasgo.connect(rasgo_api_key)
-    rasgo_transforms = utils.get_all_rasgo_transform_keyed_by_name(rasgo)
+
+    org = rasgo.get.organization()
+    if org.name != "Community":
```
griffatrasgo, line 35: should we use something like org.id == 1 (and/or make COMMUNITY_ORG_ID = 1 an int in constants.py we can ref)?

griffatrasgo: Reason being: in the future users may be able to change their org name, but we control the id (and 1 is consistent across our envs)

griffatrasgo (approved):

python/constants.py
```diff
@@ -21,3 +21,5 @@ class PyRasgoEnvironment(Enum):
 # TODO: When we support multiple DWs in the API
 # we'll need to refactor the functions that consume this
 RASGO_DATAWAREHOUSE = 'snowflake'
+
+COMMUNITY_ORGANIZATION_ID: int = 1
```
szaghikian-rasgo, line 25: I'm loving that this is going in as a var, but I wonder if it should be an enum; it's possible that we have diff org IDs we want to publish with by environment (if not now, potentially in the future). Not sure that it blocks right now, tho

szaghikian-rasgo (approved):
Looks good, left a comment about community org ID as a var vs as an enum driven by env.

python/constants.py
ChrisGriffithRASGO: May not to be bad to add if we ever need it, but as this is just our utils for now, I'm not too worried about perfect structure / plan for future as we might nix this all in a few months

PR #209, opened 2022-05-18
Added column inference

kimballh:
I added the `infer_columns` function to the package which takes the transform name and passes args to the `infer_columns` function for the respective transform if implemented in `<transform_name>/<transform_name>.py`


rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+    else:
+        return None
+
+
+def get_path(transform_name: str, dw_type: Optional[DataWarehouse] = None) -> Optional[str]:
+    root_dir = os.path.dirname(__file__)
+    if dw_type:
+        function_path = Path(root_dir, 'transforms', transform_name, dw_type.value,
+                             f'{transform_name}.py')
+        if os.path.exists(function_path):
```
ChrisGriffithRASGO, line 32:
```suggestion
        if function_path.exists():
```

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+            return str(function_path.absolute())
+    function_path = Path(root_dir, 'transforms', transform_name, f'{transform_name}.py')
+    if not os.path.exists(function_path):
+        return None
+    return str(function_path.absolute())
```
ChrisGriffithRASGO, line 37:
```suggestion
def get_path(transform_name: str, dw_type: Optional[DataWarehouse] = None) -> Optional[str]:
    transform_dir = Path(__file__).parent / "transforms" / transform_name
    function_path = transform_dir / dw_type /  f'{transform_name}.py') if dw_type else transform_dir / f'{transform_name}.py'
    if function_path.exists():
        return str(function_path)
```

The Power of Paths Compels You!

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_dw_type(dw_type: str) -> Optional[DataWarehouse]:
+    if dw_type:
+        try:
+            return DataWarehouse[dw_type.upper()]
+        except KeyError:
+            raise Exception(f'Unsupported DataWarehouse type: {dw_type}')
+    else:
+        return None
```
ChrisGriffithRASGO, line 24:
```suggestion
```

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+    remove double quotes
+    replace spaces and dashes with underscores
+    cast to upper case
+    delete anything that is not letters, numbers, or underscores
+    if first character is a number, add an underscore to the beginning
+    """
+    symbol = str(symbol)
+    symbol = symbol.strip()
+    symbol = symbol.replace(' ', '_').replace('-', '_')
+    symbol = symbol.upper()
```
ChrisGriffithRASGO, line 58:
```suggestion
    symbol = str(symbol).strip().replace(' ', '_').replace('-', '_').upper()
```

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+    symbol = re.sub('[^A-Z0-9_]+', '', symbol)
+    # if symbol is empty, at least returns '_'
+    # if starts with a decimal prefix with '_'
+    symbol = '_' + symbol if symbol[0].isdecimal() or not symbol else symbol
```
ChrisGriffithRASGO, line 62:
```suggestion
    symbol = f'_{symbol}' if not symbol or symbol[0].isdecimal() else symbol
```

You want to do the truthy check first so it won't indexerror if there isn't an index 0. 

Also use fstrings or format, and [never + to concat strings](https://codecalamity.com/stop-using-plus-signs-to-concatenate-strings/)

ChrisGriffithRASGO: if they symbol is empty, do we want to error instead?

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+        output_columns = function(
+            args=transform_args,
+            source_columns=source_columns
+        )
+        cleaned_output_columns = {}
+        for column_name, column_type in output_columns.items():
+            cleaned_output_columns[cleanse_name(column_name)] = get_dtype(column_type)
+        return cleaned_output_columns
+    else:
+        return None
```
ChrisGriffithRASGO, line 93:
```suggestion
    if not function_path:
        return None
    cleaned_source_columns = {}
    for column_name, column_type in source_columns.items():
        cleaned_source_columns[cleanse_name(column_name)] = get_dtype(column_type)
    loader = machinery.SourceFileLoader(transform_name, function_path)
    spec = util.spec_from_loader(transform_name, loader)
    module = util.module_from_spec(spec)
    loader.exec_module(module)
    function = module.infer_columns
    output_columns = function(
        args=transform_args,
        source_columns=source_columns
    )
    cleaned_output_columns = {}
    for column_name, column_type in output_columns.items():
        cleaned_output_columns[cleanse_name(column_name)] = get_dtype(column_type)
    return cleaned_output_columns
```

Generally cleaner to keep more logic at left-most level, so if easy to eliminate return cases early, best to do that

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+        loader.exec_module(module)
+        function = module.infer_columns
+        output_columns = function(
+            args=transform_args,
+            source_columns=source_columns
+        )
+        cleaned_output_columns = {}
+        for column_name, column_type in output_columns.items():
+            cleaned_output_columns[cleanse_name(column_name)] = get_dtype(column_type)
+        return cleaned_output_columns
```
ChrisGriffithRASGO, line 91:
```suggestion
        return {cleanse_name(name): get_dtype(column_type)  for name, column_type in output_columns.items()}
```

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+            return str(function_path.absolute())
+    function_path = Path(root_dir, 'transforms', transform_name, f'{transform_name}.py')
+    if not os.path.exists(function_path):
+        return None
+    return str(function_path.absolute())
+
+
+def import_from(module, name):
+    module = __import__(module, fromlist=[name])
+    return getattr(module, name)
```
ChrisGriffithRASGO, line 42: This is unused?

rasgotransforms/rasgotransforms/render.py
```diff
@@ -0,0 +1,93 @@
...
+    if function_path:
+        cleaned_source_columns = {}
+        for column_name, column_type in source_columns.items():
+            cleaned_source_columns[cleanse_name(column_name)] = get_dtype(column_type)
+        loader = machinery.SourceFileLoader(transform_name, function_path)
+        spec = util.spec_from_loader(transform_name, loader)
+        module = util.module_from_spec(spec)
+        loader.exec_module(module)
+        function = module.infer_columns
+        output_columns = function(
```
ChrisGriffithRASGO, line 84:
```suggestion
        output_columns = module.infer_columns(
```

No need for one time use variable

ChrisGriffithRASGO (changes requested):
Pretty much all suggestions, except one big omission: documentation

There isn't anything clearly stating what these are, why they are here, how to use them or build them or if they are required, anything like that.

Need to be super clear in our open source repos for anyone that stumbles unto them what they are and how to use them.

rasgotransforms/rasgotransforms/dtypes.py
```diff
@@ -0,0 +1,40 @@
+DTYPES = {
```
cnash6, line 1: We have other places in our code base where we map actual data types to simplified "rasgo" data types.  Is there any chance we could benefit from a similar treatment here?

kimballh: Yeah I think a mapping will make this easier going forward.  We do have a mapping to numeric, categorical, or datetime in the API, but we don't use it for storing column metadata; we store ansi sql types in that table so I did this to avoid making a change. I think we can push this for now and then change the mapping after we think about types some more because there have been a few points making it worth a discussion that have come up this week.

PR #212, opened 2022-05-21
This will make bigquery happier

griffatrasgo:
BigQuery and Snowflake syntax are not compatible in come cases. Making these changes to fix transform errors:
- change `::` casts to explicit `CAST` statements
- reorder `dateadd`, `datediff` args
- replace `date_part` func
- remove double quoted column names


rasgotransforms/rasgotransforms/transforms/funnel/funnel.sql
```diff
@@ -1,7 +1,7 @@
 {%- for col_name in stage_columns -%}
     SELECT
     '{{ col_name }}' AS LABEL
-    ,SUM({{ col_name }}) AS "Count"
+    ,SUM({{ col_name }}) AS LABEL_COUNT
```
griffatrasgo, line 4: @szaghikian-rasgo is this new column alias going to break anything in the UI insights chart?

szaghikian-rasgo: My understanding is that in these transforms, the UI doesn't care explicitly about column names, but I'm double checking with @phillip-rasgo to be sure

rasgotransforms/rasgotransforms/transforms/datespine/bigquery/datespine.sql
```diff
@@ -0,0 +1,40 @@
...
+{% else %}
+    {% set min_date = min_max_query_result[min_max_query_result.columns[0]][0] %}
+{% endif %}
+{% if  end_timestamp is defined %}
+    {% set max_date = end_timestamp %}
+{% else %}
+    {% set max_date = min_max_query_result[min_max_query_result.columns[1]][0] %}
+{% endif %}
+{% set row_count_query %}
+select datediff(cast('{{ min_date }}' as date), cast('{{ max_date }}' as date), {{ interval_type }})
```
szaghikian-rasgo, line 16: Isn't this built-in function `DATE_DIFF` in (BigQuery)[https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions] ?

rasgotransforms/rasgotransforms/transforms/datespine/bigquery/datespine.sql
```diff
@@ -0,0 +1,40 @@
...
+with date_spine as (
+    select
+           row_number() over (order by null) as interval_id,
+            dateadd(
+                cast('{{ min_date }}' as timestamp),
+                INTERVAL interval_id - 1 '{{ interval_type }}'
+                ) as ts_ntz_interval_start,
+            dateadd(
+                cast('{{ min_date }}' as timestamp)
+                INTERVAL interval_id '{{ interval_type }}'
```
szaghikian-rasgo, line 29:
```suggestion
            DATE_ADD(
                cast('{{ min_date }}' as timestamp)
                INTERVAL interval_id {{ interval_type }}
```

rasgotransforms/rasgotransforms/transforms/plot/bigquery/plot.sql
```diff
@@ -0,0 +1,185 @@
+{%- set axis_type_dict = get_columns(source_table) -%}
```
szaghikian-rasgo, line 1: Same possible `get_columns` macro issue

rasgotransforms/rasgotransforms/transforms/describe/bigquery/describe.sql
```diff
@@ -0,0 +1,42 @@
+{%- set names_types_list = get_columns(source_table) -%}
```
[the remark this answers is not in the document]
szaghikian-rasgo: 99.9999% sure it's stored as lowercase strings, which won't return with this query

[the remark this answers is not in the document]
griffatrasgo: Great callout. We will 100% need to drop the .upper() from the input vars.

For BQ, the column names will return exactly as they are created in the DW. We should not do any case trapping in or out of the DW and just let them flow as stored.

griffatrasgo:
> I'm really worried about the interplay with our API macros, specifically `get_columns`, which I'm pretty sure is going to not return anything right now, since table catalog, schema, and table names will all be stored in lowercase in BQ's information schema

This is a great callout. We will need to change this function in each of the consumers (rql, pyrasgo, api)

PR #221, opened 2022-05-26
new block for value in list

szaghikian-rasgo:
PR fixing issue with value in list comparison operator

### DIFF
--- a/rasgotransforms/rasgotransforms/transforms/filter/filter.sql
+++ b/rasgotransforms/rasgotransforms/transforms/filter/filter.sql
@@ -16,6 +16,8 @@ FROM {{ source_table }}
 {%- else -%}
     {%- if filter_block['operator'] == 'CONTAINS' -%}
 {{ filter_block['operator'] }}({{ filter_block['columnName'] }}, {{ filter_block['comparisonValue'] }})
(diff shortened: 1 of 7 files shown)


rasgotransforms/rasgotransforms/transforms/filter/filter.sql
```diff
@@ -16,6 +16,8 @@ FROM {{ source_table }}
 {%- else -%}
     {%- if filter_block['operator'] == 'CONTAINS' -%}
 {{ filter_block['operator'] }}({{ filter_block['columnName'] }}, {{ filter_block['comparisonValue'] }})
+    {%- elif filter_block['operator'] == 'IN' -%}
+        {{ filter_block['columnName'] }} {{ filter_block['operator'] }} ({{ filter_block['comparisonValue'] }})
```
ChrisGriffithRASGO, line 20:
```suggestion
    {%- elif filter_block['operator'] == 'IN' -%}
        {{ filter_block['columnName'] }} IN ({{ filter_block['comparisonValue'] }})
```
For sections like these curious if it might be more readable to just put the operator in them directly. (Same is true for the CONTAINS as well.)

ChrisGriffithRASGO (approved):

PR #227, opened 2022-06-01
cast user input cols to uppercase

kimballh:


griffatrasgo (commented):
Agree with the fix. Can we break this template into a Snowflake & BQ version. And only apply this fix in the Snowflake version?

griffatrasgo (approved):

PR #231, opened 2022-06-03
Auto publish transforms to S3

cnash6:
Add an action to auto-publish transforms to S3 so SG5K can consume

### DIFF
--- a/.github/workflows/publish_transforms_to_s3.yaml
+++ b/.github/workflows/publish_transforms_to_s3.yaml
@@ -33,4 +33,3 @@ jobs:
 
       - name: Publish Transforms to S3
         run: python python/publish_transforms_to_s3.py
-
--- a/.github/workflows/publish_transforms_to_s3.yaml
+++ b/.github/workflows/publish_transforms_to_s3.yaml
@@ -1,0 +1,36 @@
+# Publish all Transforms in this Repo to S3
+# Run job each time something is pushed/committed to remote 'main' branch
+name: Publish Transforms to S3
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish-accelerators:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+
+    container:
+      image: "python:3.7"
+
+    env:
+      PYTHONPATH: /__w/RasgoTransforms/RasgoTransforms
+      AWS_ACCESS_KEY_ID: ${{ secrets.S3_UPLOAD_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_UPLOAD_SECRET_ACCESS_KEY }}
+
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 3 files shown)


python/publish_transforms_to_s3.py
```diff
@@ -0,0 +1,38 @@
...
+Python Script to Publish all Transforms (for every DW) to S3
+"""
+import json
+import rasgotransforms
+import boto3
+
+
+def to_camel_case(snake_str):
+    components = snake_str.split('_')
+    return components[0] + ''.join(x.title() for x in components[1:])
```
ChrisGriffithRASGO, line 11:
```suggestion
    return f"{components[0]}{''.join(x.title() for x in components[1:])}"
```

python/publish_transforms_to_s3.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_some_transforms(dw_type: str):
+    return [
+        {**x.__dict__, **{'dw_type': dw_type}}
```
ChrisGriffithRASGO, line 16:
```suggestion
        {**x.__dict__, 'dw_type': dw_type}
```

Why all the explodin'?

python/publish_transforms_to_s3.py
```diff
@@ -0,0 +1,38 @@
...
+        for x in rasgotransforms.serve_rasgo_transform_templates(datawarehouse=dw_type)
+    ]
+
+
+def publish_transforms() -> None:
+    """
+    Build the json and push it to S3
+    """
+    # get ALL the transforms for each DW Type in the DataWarehouse Enum
+    raw_transforms = [x for y in [get_some_transforms(x.name) for x in rasgotransforms.main.DataWarehouse] for x in y]
```
ChrisGriffithRASGO, line 26: Please don't reuse the same variable name three times in the same list comp

python/publish_transforms_to_s3.py
```diff
@@ -0,0 +1,38 @@
+"""
+Python Script to Publish all Transforms (for every DW) to S3
+"""
+import json
+import rasgotransforms
+import boto3
```
ChrisGriffithRASGO, line 6:
```suggestion
import json

import boto3

import rasgotransforms
```

Standard lib 

External 

Model

.github/workflows/publish_transforms_to_s3.yaml
```diff
@@ -0,0 +1,35 @@
...
+      image: "python:3.7"
+
+    env:
+      PYTHONPATH: /__w/RasgoTransforms/RasgoTransforms
+      AWS_ACCESS_KEY_ID: ${{ secrets.S3_UPLOAD_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_UPLOAD_SECRET_ACCESS_KEY }}
+
+
+    steps:
+      - uses: actions/checkout@v2
```
ChrisGriffithRASGO, line 27:
```suggestion
      - uses: actions/checkout@v3
```

PR #233, opened 2022-06-09
Make case sensitivity false by default on Drop Columns

cnash6:
Proposal: Case sensitivity shouldn't matter for most column comparisons.  We should compare case insensitive by default. If we need to compare _using_ case sensitivity then that should be an active choice a user makes. 

Separate proposal: This distinction might not matter to people for now, so we could leave this variable off of the list of arguments we expose to users and have it be a secret internal variable that we use in the future. 

other separate proposal: We could make secret internal variables a real thing....


griffatrasgo (commented):
Love this pattern! You're very smart man. Two small usability suggestions below. Then... ship it!

rasgotransforms/rasgotransforms/transforms/drop_columns/drop_columns.yaml
```diff
@@ -16,6 +16,10 @@ arguments:
     type: column_list
     description: A list of the columns from the dataset you want to drop. Any columns not in the exclude_cols list will be kept.
     is_optional: true
+  case_sensitive_exclude:
```
griffatrasgo, line 19: could we name this something generic that can be used across templates. I can instantly see this expanding to all templates that invoke `get_columns`

Options?
case_sensitive_match
case_sensitive
case_sensitive_column_names

rasgotransforms/rasgotransforms/transforms/drop_columns/drop_columns.sql
```diff
@@ -12,8 +12,13 @@ FROM {{source_table}}
 
 {%- if exclude_cols is defined -%}
 {%- set source_col_names = get_columns(source_table) -%}
-{%- set new_columns = source_col_names | reject('in', exclude_cols) -%}
 
+    {%- if not case_sensitive_exclude -%}
```
griffatrasgo, line 16: Could we write this in a way that defaults to True? That way users only need to pass the param if they want to turn ON case sensitivity (which we'd be betting is the less frequently occurring case)

kimballh: I agree with this and I think the way it is written it is case insensitive by default which is what we want, but the name might make more sense if it's just 'case_sensitive'.

cnash6:
Updated with @griffatrasgo's notes. We shouldn't change the return case of the inputs!

PR #234, opened 2022-06-10
Kh/render metrics

kimballh:
This is the transform used to calculate metrics as timeseries aggregation, accepting the metric definition fields as input.

### DIFF
--- a/rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
+++ b/rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
@@ -1,0 +1,56 @@
+name: metric
+type: insight
+tags:
+  - insight
+description: Calculate metric values given a metric definition
+arguments:
+  timestamp:
+    type: column
+    description: The date/datetime column by which the metric will be aggregated
+  target_sql:
+    type: column
+    description: The target column used to calculate the metric, can also be a sql string '(e.g. cast(target_column as string)')
+  aggregation_type:
+    type: string
+    description: the Aggregation method used to calculate the metric (e.g. 'sum', 'count_distinct', 'max')
+  time_grain:
+    type: string
+    description: The time grain (day, week, month, quarter, year) used to create the datespine that the metric is aggregated by
+  dimensions:
+    type: column_list
+    description: List of columns used to group by in metric aggregation
+    is_optional: true
+  filters:
+    type: list
+    description: List of objects containing filters used to filter the source data. A filter object must contain the fields 'field', 'operator', and 'value'. String values should be wrapped in single quotes.
+    is_optional: true
+  alias:
+    type: string
+    description: The name of the output metric value column
+    is_optional: true
+
+example_code: |
+  ds = rasgo.get.dataset(id)
+
+  ds2 = ds.metric(
+      source_table="DBTTEST.TUTORIAL.ORDERS",
+      timestamp='ORDERDATE',
+      target_sql=SALESAMOUNT',
+      aggregation_type='sum',
+      alias='weekly_sales',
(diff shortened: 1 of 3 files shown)


rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
```diff
@@ -0,0 +1,56 @@
+name: metric
+type: insight
+tags:
+  - insight
+description: Calculate metric values given a metric definition
+arguments:
+  timestamp:
```
cnash6, line 7: Is naming a column argument "timestamp" confusing?  Would something like "time_column" make more sense?

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
```diff
@@ -0,0 +1,56 @@
...
+    type: column
+    description: The date/datetime column by which the metric will be aggregated
+  target_sql:
```
cnash6, line 10: `target_sql` can also be just a column name? maybe `target_expression`? or just `target`?

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
```diff
@@ -0,0 +1,56 @@
...
+    description: the Aggregation method used to calculate the metric (e.g. 'sum', 'count_distinct', 'max')
+  time_grain:
+    type: string
+    description: The time grain (day, week, month, quarter, year) used to create the datespine that the metric is aggregated by
+  dimensions:
+    type: column_list
+    description: List of columns used to group by in metric aggregation
+    is_optional: true
+  filters:
+    type: list
```
cpdough, line 24: should we use `    type: filter_list` here so that the UI input will match? in case users want to edit a metric rendering step

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
```diff
@@ -0,0 +1,56 @@
...
+    type: column
+    description: The date/datetime column by which the metric will be aggregated
+  target_sql:
+    type: column
+    description: The target column used to calculate the metric, can also be a sql string '(e.g. cast(target_column as string)')
+  aggregation_type:
+    type: string
+    description: the Aggregation method used to calculate the metric (e.g. 'sum', 'count_distinct', 'max')
+  time_grain:
+    type: string
```
cpdough, line 17: i think we support `    type: date_part` for this

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
```diff
@@ -0,0 +1,56 @@
...
+description: Calculate metric values given a metric definition
+arguments:
+  timestamp:
+    type: column
+    description: The date/datetime column by which the metric will be aggregated
+  target_sql:
+    type: column
+    description: The target column used to calculate the metric, can also be a sql string '(e.g. cast(target_column as string)')
+  aggregation_type:
+    type: string
```
cpdough, line 14: `    type: agg`

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
kimballh: Yeah I like target_expression better.

rasgotransforms/rasgotransforms/transforms/metric/metric.yaml
kimballh: The only issue is that this template will only support 'day', 'week', 'month', 'quarter', and 'year' while `date_part` has a bunch of other date part options.

PR #235, opened 2022-06-13
Jinja Rendering

kimballh:
This adds the `render` function to the RasgoTransforms package, which takes a template source code, args, dw_type, and source columns to render the sql. It also adds the `transform` module which has methods for each of the transforms to render (loaded at runtime). If calling the transform class method, or the render function, the `run_query` and `get_columns` methods can be passed for use by the transform templates.


rasgotransforms/rasgotransforms/jinja.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Source columns can be a mapping of table names to columns (e.g. {table_name: {column_name: column_type}})
+    or a function which returns a mapping of columns names to column types given a table fqtn
+    (e.g. get_columns(fqtn): return {column_name: column_type}
+
+    Run Query is a function which takes arbitrary sql and executes it agains the data warehouse. The data must be
+    returned as a Pandas dataframe
+    """
+    arguments['source_table'] = source_table
+    if source_columns and type(source_columns) is dict:
```
cnash6, line 43: Why allow dict _or_ a function? Why not just keep it simple and make it always a function (that just returns the correct dict)

For transparency, this is what I was thinking: 
https://github.com/rasgointelligence/render-transforms/blob/master/lambda_function.py#L46

rasgotransforms/rasgotransforms/jinja.py
```diff
@@ -0,0 +1,60 @@
...
+    raise Exception(message)
+
+
+def render(
+    source_code: str,
+    source_table: str,
+    arguments: dict,
+    source_columns: Optional[Union[Dict[str, Dict[str, str]], Callable]] = None,
+    run_query: Optional[Callable] = None,
+) -> str:
```
cnash6, line 30: Instead of expecting discrete function overrides, why not accept a dict of overrides that we can merge onto the local `template_fns` that way clients can choose which methods to overwrite (or add their own where necessary?)

rasgotransforms/rasgotransforms/transforms.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self):
+        transforms_dir = Path(os.path.dirname(__file__), 'transforms')
+        for transform_path in transforms_dir.rglob('*/*.yaml'):
+            with open(transform_path) as fp:
+                transform_config = yaml.safe_load(fp)
+            setattr(self.__class__, transform_config['name'], _get_render_method(transform_config))
+
+
+def _gen_method_signature(udt_func: Callable, transform_config: Dict) -> inspect.Signature:
```
ChrisGriffithRASGO, line 24: Please don't do module level `_` underscore started methods. Use `__all__ = ["Transforms"]` at the top instead. 

More details on why here: https://rasgoteam.slack.com/archives/C03A8E9PZ5G/p1654612707517819?thread_ts=1654525069.737569&cid=C03A8E9PZ5G

rasgotransforms/rasgotransforms/transforms.py
```diff
@@ -0,0 +1,113 @@
...
+from .main import DataWarehouse
+
+
+class Transforms(object):
+    """
+    Dynamically adds class methods which render each transform in the package
+    """
+
+    def __init__(self):
+        transforms_dir = Path(os.path.dirname(__file__), 'transforms')
```
ChrisGriffithRASGO, line 17:
```suggestion
        transforms_dir = Path(__file__).parent / 'transforms'
```

Can remove the `os` module and just switch to using `pathlib` throughout the file

rasgotransforms/rasgotransforms/transforms.py
```diff
@@ -0,0 +1,113 @@
...
+    render_transform.__signature__ = _gen_method_signature(render_transform, transform_config)
+    render_transform.__doc__ = _gen_method_docstring(transform_config)
+    return render_transform
+
+
+def _get_transform_path(name: str, dw_type: DataWarehouse):
+    """
+    Returns the default or data warehouse specific template file given the transform name
+    """
+    root_dir = os.path.dirname(__file__)
```
ChrisGriffithRASGO, line 106:
```suggestion
    root_dir = Path(__file__).parent
```

rasgotransforms/rasgotransforms/transforms.py
```diff
@@ -0,0 +1,113 @@
...
+    if dw_type:
+        function_path = Path(root_dir, 'transforms', name, dw_type.value, f'{name}.sql')
+        if function_path.exists():
+            return str(function_path.absolute())
+    function_path = Path(root_dir, 'transforms', name, f'{name}.sql')
+    if os.path.exists(function_path):
```
ChrisGriffithRASGO, line 112:
```suggestion
    if function_path.exists():
```

rasgotransforms/rasgotransforms/jinja.py
```diff
@@ -0,0 +1,60 @@
...
+import re
+
+JINJA_ENV = Environment(extensions=['jinja2.ext.do', 'jinja2.ext.loopcontrols'])
+
+
+def cleanse_template_symbol(symbol: str) -> str:
+    symbol = str(symbol)
+    symbol = symbol.strip()
+    symbol = symbol.replace(' ', '_').replace('-', '_')
+    symbol = symbol.upper()
```
griffatrasgo, line 13: probably need to drop the .upper() here

rasgotransforms/rasgotransforms/jinja.py
griffatrasgo: +1

rasgotransforms/rasgotransforms/jinja.py
griffatrasgo: +1, but also why accept source_columns at all? Just let them pass in get_columns as one of the functions in the override dict as suggested above.
We're going to find 10 more macros we think every transform needs if this package gets wider adoption. We should draw a line early and say all macros (functions) get passed in a single arg to keep the `render` function signature clean

kimballh: I will add the override dict for other macros. My thoughts for adding the source_columns arg and accepting a dict or a function were that if we had a function which returns the rendered sql and the inferred output columns then we could just pass those output columns into the function again to render another transform, so chaining lots of transforms is really easy like this:

```
transform1_sql, output_columns = render(..., infer_columns=True)
transform2_sql, output_columns = render(..., source_columns=output_columns, infer_columns=True)
```

The reason I take a function and not just a dict is so that our api or other clients that work similarily can pass in a 'get_columns' function.

If accepting the source_columns like this doesn't seem useful, I can take it out.

rasgotransforms/rasgotransforms/jinja.py
kimballh: Should I leave `run_query` as it's own arg or just take it in the override dict? My thought was that if run_query was provided, any other macros could be defined in the jinja templates themselves. The more macros we allow to live outside of the jinja template, the more we require our clients to implement and pass in. 

If the only macro that we externally define here is `run_query` and force the transforms to implement any other macros inline with jinja, then it makes it much easier for clients to integrate. They only need to implement `run_query` which should be the only client specific macro, defining how they execute sql against the DW.

rasgotransforms/rasgotransforms/render/transforms.py
```diff
@@ -0,0 +1,120 @@
...
+
+__all__ = ["Transforms"]
+
+
+class Transforms(object):
+    """
+    Dynamically adds class methods which render each transform in the package
+    """
+
+    def __init__(self, dw_type: str, run_query: Optional[Callable] = None):
```
griffatrasgo, line 19: is there a reason we want to break out `run_query` here vs accepting a dict of `global_overrides` (of which run_query would be one)?

rasgotransforms/rasgotransforms/render/environment.py
```diff
@@ -0,0 +1,76 @@
...
+        source_table: str,
+        arguments: dict,
+        source_columns: Optional[Dict[str, Dict[str, str]]] = None,
+        override_globals: Optional[Dict[str, Callable]] = None,
+    ) -> str:
+        """
+        Given the source code of a jinja template, its arguments, and metadata about the source table,
+        this renders the transform sql.
+
+        Source columns can be a mapping of table names to columns (e.g. {table_name: {column_name: column_type}})
```
griffatrasgo, line 45: It looks like we're only accepting a dict now (which I agree is the correct answer). May just need to update this comment to reflect that

rasgotransforms/rasgotransforms/render/transforms.py
kimballh: Since this environment inherits from a jinja environment, globals can still be passed in here as kwargs. The reason that `run_query` is pulled out is because it is the only macro which does not have a default implementation; every other macro that our transforms use are defined in this environment.

griffatrasgo (approved):

PR #236, opened 2022-06-14
Clean up transform naming

cpdough:
Still need to update accelerators to match


cpdough:
this breaks b2b_funnel because it removes support for multi_join and multi_union.

cpdough:
**summary of changes**
deleted:

- multi union
- multi join
- simple join

renamed:
- split
- impute
- dedupe
- outliers

PR #251, opened 2022-06-23
improved datespine

kimballh:
Big improvements for datespine:
- Big query version didn't work at all before, now it works
- num_rows is calculated from start and stop dates in the jinja with python instead of using runquery
- week periods now go back to the first day of the calendar week instead of just 7 days before whatever the min_date is
- added support for quarter time grain
- Made same changes to datespine_groups


cnash6:
Looks good to me, but we should return meaningful exceptions (like in one_hot or pivot) when we can't build the template because we don't have optional params or run_query returns nothing.

kimballh:
@cnash6 I added in the error message.

PR #259, opened 2022-06-29
Verify release tag matches version in version.py

kimballh:
This would ensure that the release tag and the pypi version are in sync


kimballh:
@ChrisGriffithRASGO I'm not really sure what the best way to test this would be, do you know how to test this without merging it to main and then creating a release?

PR #265, opened 2022-07-11
Metric group by bug fixes

kimballh:
- Better support for  numerical dimensions for metrics
- apply metric filters when getting top groups as well as filter for given start_date


cpdough:
so you didn't have to remove the 'Other' from the list of distinct values to get it working? that was tripping me up yesterday

kimballh:
It doesn't matter if we leave 'Other' in the list because it's just checking if the value is in the list.

PR #270, opened 2022-07-14
Fix broken bigquery transforms

szaghikian-rasgo:
Notes about changes:
1. Aggregate
- Killed entropy in the UI codebase
- Killed listagg options in transforms
2. Cast
- float to float64
- number to numeric
3. Clean
- Impute is broken in Snowflake and BigQuery - if the source value is a non-string, the default impute being a string will create problems. Instead, make the default value `null` so that we don't have type mismatches in the `COALESCE` args


rasgotransforms/rasgotransforms/transforms/clean/clean.sql
```diff
@@ -11,7 +11,7 @@
 {%- macro get_select_column(name, col) -%}
     {%- set source_col = 'cast(' + name + ' as ' + col.type + ')'  if col.type is defined else name-%}
     {% set output_col_name = name if col.name is not defined else cleanse_name(col.name) %}
-    {%- set impute_expression = '' -%}
+    {%- set impute_expression = none -%}
```
griffatrasgo, line 14: (asking to learn, not bc i know anything) jinja wants none instead of None?

griffatrasgo (approved):
Well done! Impossible to review jinja, but I'm sure you tested twice so... roll tide.

rasgotransforms/rasgotransforms/transforms/clean/clean.sql
szaghikian-rasgo: No actually this needs to be NULL since it's going straight into our SQL string

rasgotransforms/rasgotransforms/transforms/aggregate/aggregate.sql
```diff
@@ -4,10 +4,14 @@ SELECT
 {%- endfor -%}
 
 {%- for col, aggs in aggregations.items() %}
-        {%- set outer_loop = loop -%}
+    {%- set outer_loop = loop -%}
     {%- for agg in aggs %}
-    {{ agg }}({{ col }}{{ ", '|'" if agg|lower == 'listagg' else ""}}) as {{ col + '_' + agg }}{{ '' if loop.last and outer_loop.last else ',' }}
+        {%- if ' DISTINCT' in agg|upper %}
+            {{ agg|replace(" DISTINCT", "") }}(DISTINCT {{ col }}) as {{ col ~ '_' ~ agg|replace(" DISTINCT", "") ~ 'DISTINCT'}}{{ '' if loop.last and outer_loop.last else ',' }}
```
kimballh, line 10:
```suggestion
            {{ agg|upper|replace(" DISTINCT", "") }}(DISTINCT {{ col }}) as {{ col ~ '_' ~ agg|upper|replace(" DISTINCT", "") ~ 'DISTINCT'}}{{ '' if loop.last and outer_loop.last else ',' }}
```

Do we check anywhere that the methods passed in are uppercase? This might be unecessary if so.

rasgotransforms/rasgotransforms/transforms/lead/bigquery/lead.sql
```diff
@@ -0,0 +1,19 @@
+{%- if partition is not defined or partition|length == 0 -%}
+{%- set partition = ["NULL"]-%}
+{%- endif -%}
+{%- if order_by is not defined or order_by|length == 0 -%}
+{%- set order_by = ["NULL"]-%}
+{%- endif -%}
+{%- for amount in amounts -%}
+    {%- if amount < 0 -%}
+        {{ raise_exception('BigQuery cannot use negative values for a lead function. Please utilize lead for backwards looking windows.') }}
```
kimballh, line 9:
```suggestion
        {{ raise_exception('BigQuery cannot use negative values for a lead function. Please utilize lag for backwards looking windows.') }}
```

kimballh (commented):
I tested them with the one table we have and they worked for me. Aggregate got way more complicated, but bigquery just sucks like that I guess.

PR #293, opened 2022-08-02
it cant be this easy?

szaghikian-rasgo:
Big problem we used to have in BigQuery pivot - we weren't aliasing the column names as we pivoted them out, because the template wasn't constructed in a friendly way to allow us to use `cleanse_name` (or so we thought). Turns out, you can functionally alias column names in the `IN()` clause, meaning we can rock with `cleanse_name` in this template and stop seeing bugs.


griffatrasgo (approved):

kimballh (approved):
Assuming you've tested it, LGTM!

PR #302, opened 2022-08-11
Kh/ratio metrics

kimballh:
This has a new template for ratio metrics as well as changes to the render environment and additional macros which will make developing templates easier.

Ratio Metrics
- Same params as metric template, with a few changes:
  - `target_expression` is not just a column, but an expression made up of metrics 
    Example: `SALES_REVENUE / SALES_VOLUME`
  - `metrics` parameter must be passed in which is a list of objects that define metrics as a column name, agg method, and alias. Each metric used in the `target_expression` must be in this list of metrics, the `target_expression` refers to metrics by their alias. 
    Example:
    ```
    [
      {'alias': 'SALES_REVENUE', 'column': 'SALESAMOUNT', 'agg_method': 'SUM'},
      {'alias': SALES_VOLUME, 'column': 'ORDERNUMBER', 'agg_method': 'COUNT DISTINCT'},
    ]
    ```

Rendering Improvements
- Macros are defined in `rasgotransforms/macros` 
  - Macros are loaded with a custom `RasgoLoader` which loads macros according to the `RasgoEnvironment.dw_type` property
  - This change requires the `dw_type` to be set when the `RasgoEnvironment` is initialized, so it will cause a breaking change to our clients which currently don't pass it in
- White space is managed much better now so we don't need to add whitespace modifiers to our blocks ({%`-`...)  
- Added [sqlfmt](https://github.com/tconbeer/sqlfmt) to precommit which actually does a beautiful job at formatting jinja sql


rasgotransforms/rasgotransforms/snippets/alias.sql
```diff
@@ -1 +1 @@
-AS {{ alias }}
\ No newline at end of file
+as {{ alias }}
```
griffatrasgo, line 1: files in the snippets folder shouldn't have a line break at end of the file. i'm guessing the new sqlformat hook is doing this? Any way we can exclude this folder from the formatting?

kimballh: Yeah I am pulling the formatter into it's own pr, so I reverted those changes. I'll be sure to exclude snippets in that one though.

rasgotransforms/setup.py
```diff
@@ -42,6 +42,8 @@
             'transforms/**/**/*.sql',
             'transforms/**/**/*.py',
             'snippets/*.sql',
+            'macros/*.sql',
```
griffatrasgo, line 45: 👍

PR #303, opened 2022-08-12
Add sqlfmt to precommit

kimballh:


griffatrasgo:
Can you share the pros/cons of this package and why we should pick it vs the sqlparse package we already use in the API?

PR #304, opened 2022-08-18
all or any

cpdough:


rasgotransforms/rasgotransforms/transforms/filter/filter.sql
```diff
@@ -10,7 +10,7 @@ SELECT *
 FROM {{ source_table }}
 {% for filter_block in items %}
 {%- set oloop = loop -%}
-{{ 'WHERE ' if oloop.first else ' AND ' }}
+{{ 'WHERE ' if oloop.first else ' ' ~ all_or_any ~ ' ' }}
```
sam-rasgo, line 13: I was actually going to store this in the filter object itself so that way we can grab it from the filter object and we won't have to pass around a new variable

sam-rasgo, line 13: Scratch any comment you've seen so far. I need to make sure it'll work for both preview filters and transform filters

sam-rasgo: Instead of all_or_any it will be:
`filter_block['compoundBoolean']`

PR #305, opened 2022-08-23
added compoundBoolean to filter transform

kimballh:


rasgotransforms/rasgotransforms/transforms/filter/filter.yaml
```diff
@@ -19,5 +19,5 @@ example_code: |
   ds2.preview()
 
   # full filtering with a column, operator, and comparison value
-  ds3 = ds.filter(items=[{'columnName':'PRODUCTKEY', 'operator':'>', 'comparisonValue':'101'}])
+  ds3 = ds.filter(items=[{'columnName':'PRODUCTKEY', 'operator':'>', 'comparisonValue':'101', 'compoundBoolean': 'OR'}])
```
cpdough, line 22: looks like compoundBoolean is optional to include in the items dict, right? let's leave it out of the code example if so since you're unlikely to use that if writing code in pyrasgo

PR #314, opened 2022-09-02
Jinja testing

kimballh:
Unit tests for transforms and macros using pytest

Tests:
- Each transform and macro template file are loaded and parsed, ensuring valid jinja syntax
- Checks that the variables declared in the yaml match the variables used in the template
  - throws a warning for undeclared variables, fails if declared variables aren't used at all
- Checks that any macro or template referenced in a template exists


griffatrasgo (changes requested):
There appear to be changes unrelated to testing in this PR. Can you split those changes into a separate PR so we can focus on one thing at a time?

kimballh:
@griffatrasgo I apologize, all the other changes were in response to these tests failing, but I will put them in a separate PR.

griffatrasgo:
> all the other changes were in response to these tests failing

sounds like your tests are working then 😁 👍

rasgotransforms/rasgotransforms/tests/valid_jinja_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+LOGGER = logging.getLogger(__name__)
+MACROS = get_all_macros()
+transforms = get_all_transform_templates()
+
+
+class TestJinja:
+    @pytest.mark.parametrize("transform", transforms, ids=[t.name for t in transforms])
+    def test_transforms_jinja(self, transform: TransformTemplate):
+        environment = RasgoEnvironment(dw_type='snowflake', run_query=None)
```
griffatrasgo, line 19: Should we do a parallel run with `environment = RasgoEnvironment(dw_type='bigquery', run_query=None)` ?

rasgotransforms/rasgotransforms/transforms/plot/plot_test.py
```diff
@@ -0,0 +1,215 @@
+import unittest
```
griffatrasgo, line 1: Organization recommendation: we may want move this and future test files into the `/tests/` folder.

Since this is a public repo, there's value in keeping the tests separate from the transform library

rasgotransforms/rasgotransforms/testing_utils.py
```diff
@@ -0,0 +1,131 @@
+from pathlib import Path
```
griffatrasgo, line 1: Organization recommendation: we may want move this file into the `/tests/` folder.

Since this is a public repo, there's value in keeping the test functions separate from the transform library

rasgotransforms/rasgotransforms/testing_utils.py
```diff
@@ -0,0 +1,131 @@
...
+CREDS = {
+    "account": SNOWFLAKE_ACCOUNT,
+    "warehouse": SNOWFLAKE_WAREHOUSE,
+    "user": SNOWFLAKE_USER,
+    "password": SNOWFLAKE_PASSWORD,
+    "role": SNOWFLAKE_ROLE,
+}
+
+
+def query_into_dataframe(query) -> pd.DataFrame:
```
griffatrasgo, line 37: I don't see these called anywhere in the test scripts (which is good), assuming they're called by a notebook or file that's no git tracked in this repo.
If that's the case, you may want to add a comments block at the top of this file warning fools not to add live DW connections to tests.

rasgotransforms/rasgotransforms/tests/valid_jinja_test.py
kimballh: Good catch, I fixed this to pass in dw_type

rasgotransforms/rasgotransforms/testing_utils.py
kimballh: The Black formatter yelled at me because everything in the `tests` dir should be a test ending in `_test.py`

rasgotransforms/rasgotransforms/testing_utils.py
kimballh: In `plot_test.py` I included a bunch of tests that I was running locally. These tests are really only useful if they execute the sql, so I added the ability to do so if the snowflake creds are included in .env. I marked this tests as 'skip' because they will fail if running in a github action. I can add some documentation about writing local tests like this, what are your thoughts? I could also just remove the tests, but I thought there was value in adding them even though they need to be run locally with snowflake creds setup.

rasgotransforms/rasgotransforms/testing_utils.py
cnash6: just call that file `test.py` then?

rasgotransforms/rasgotransforms/testing_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        if not artifact_subdir.exists():
+            artifact_subdir.mkdir(parents=True)
+        with open(artifact_subdir / "args.json", "w") as fp:
+            json.dump(args, fp)
+        with open(artifact_subdir / "query.sql", "w") as fp:
+            fp.write(sql)
+        output.to_csv(artifact_subdir / "output.csv", index=False)
+
+
+def _get_templates(transform_name: str) -> dict:
```
cnash6, line 81: is there a reason this is `_`?  We're trying to use the `all` annotation if you want to hide certain functions externally

griffatrasgo (approved):

PR #316, opened 2022-09-06
'All' time time grain fix

kimballh:
The 'all' time grain is broken when adding dimensions due to the column name being assigned incorrectly in the macro, this fixes it. It also fixes a jinja syntax bug when pivoting out the dimensions for bigquery.


griffatrasgo (approved):
Much easier to review this way. TY

PR #318, opened 2022-09-14
fixed big query plot bugs and added unit tests

kimballh:
Fixes:
- cast to decimal instead of float for buckets in categorical axis
- add source missing columns to source query for multiple dimensions
- Adds bigquery local tests for plot


rasgotransforms/rasgotransforms/macros/bigquery/pivot.sql
```diff
@@ -19,18 +19,18 @@ with
...
                 {% for val in distinct_values %}
-                {% set column_name = metric_name + '_' + (val|string) %}
+                {% set column_name = metric_name + '_' + cleanse_name(val|string) %}
                 {% do column_names.append(column_name) %}
                 {% if val is string %}
-                '{{ val }}'
+                '{{ val }}' {{ cleanse_name(val) }}
                 {% else %}
-                {{ val }}
+                {{ val }} {{ cleanse_name(val) }}
```
cnash6, line 33: Does this alias the 'val'?

kimballh: Yes, because the 'val' will become a column and we want the column name to be sql friendly despite what the value is

PR #319, opened 2022-09-15
error if no results

cnash6:


rasgotransforms/rasgotransforms/transforms/plot/plot.sql
```diff
@@ -72,6 +72,9 @@
     source_table=source_table,
     filters=distinct_vals_filters
 ) | from_json %}
+{% if not distinct_values is defined or not distinct_values %}
+    {{ raise_exception('This query produces and empty result set') }}
```
kimballh, line 76:
```suggestion
{% if distinct_values is not defined or not distinct_values %}
    {{ raise_exception('This query produces an empty result set') }}
```

kimballh: This looks like it should work

PR #330, opened 2022-10-31
change filter macro to take snake and camel case

kimballh:
This is needed because sometimes the UI passes the transform args directly, which may include metric definitions that will come camel case, but other times metric definitions passed as args from the API are snake case

### DIFF
--- a/rasgotransforms/rasgotransforms/macros/filter.sql
+++ b/rasgotransforms/rasgotransforms/macros/filter.sql
@@ -19,18 +19,18 @@
     {% if filter is not string and filter is not mapping %}
     {% set filter = dict(filter) %}
     {% endif %}
-    {% if 'column_name' in filter %}
-        {% do filter.__setitem__('columnName', filter.column_name) %}
+    {% if 'columnName' in filter %}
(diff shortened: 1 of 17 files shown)


rasgotransforms/rasgotransforms/macros/filter.sql
```diff
@@ -19,6 +19,12 @@
     {% if filter is not string and filter is not mapping %}
     {% set filter = dict(filter) %}
     {% endif %}
+    {% if 'column_name' in filter %}
```
griffatrasgo, line 22: I would invert this logic and replace all instance of camelCase vars with snake_case vars in this package. This is an open-source python package, and as such we should be unapologetic about following python standards. We should use snake in every macro and template and require consumers (our UI) to make case translations when passing in values

rasgotransforms/rasgotransforms/version.py
```diff
@@ -1,4 +1,4 @@
 """
 Package version for pypi
 """
-__version__ = "2.2.14"
+__version__ = "2.2.15a2"
```
ChrisGriffithRASGO, line 4:
```suggestion
__version__ = "2.3"
```

PR #331, opened 2022-11-16
Secondary Calculations

kimballh:
Adds secondary calculations to metric and plot transform.

Secondary calculations (see [dbt](https://github.com/dbt-labs/dbt_metrics#secondary-calculations)):
- period over period
- period to date
- rolling
- prior

Secondary calculations are calculated wherever metrics are rendered (`plot` and `metric` transforms) by using the new `render_secondary_calculations` macro. This macro imports each of the secondary calculations as their own individual macros and renders them according to the `calc_config` property passed into the transform args.

Another small change in this pr is adding the `dw_type()` function to the environment, so we can use it anywhere we are writing jinja to check the dw type. This is really nifty to do stuff like this:
```
    {% if dw_type() == 'bigquery' %}
        date_trunc(period_min, {{ period }})
    {% else %}
        date_trunc('{{ period }}', period_min)
    {% endif %}
```


rasgotransforms/rasgotransforms/render/environment.py
```diff
@@ -44,6 +44,7 @@ def rasgo_globals(self):
             "cleanse_name": cleanse_template_symbol,
             "raise_exception": raise_exception,
             "itertools": {"combinations": combinations, "permutations": permutations, "product": product},
+            "dw_type": lambda: self.dw_type.value,
```
griffatrasgo, line 47: why the lambda here?

kimballh: This is the list of globals/functions that are available in the jinja environment, so this makes `dw_type()` available in any over our transforms to get the dw_type. There is an example of this in the pr description.

PR #336, opened 2022-12-08
Move filters to function

griffatrasgo:
This PR moves the filters macros into global functions


rasgotransforms/rasgotransforms/render/environment.py
```diff
@@ -73,40 +90,83 @@ def render(
...
+
+        # Handle override operators
+        if operator not in ALLOWED_OPERATORS:
+            raise_exception(f"operator {operator} is not supported")
+        if operator == "CONTAINS":
+            operator = "LIKE"
+
+        compound_boolean = "" if filter_string == "" else compound_boolean
+        filter_string += f"{compound_boolean} {column_name} {operator} {comparison_value} "
+
```
kimballh, line 160:
```suggestion
    filter_string = "TRUE"
    compound_boolean = "AND"
    for fil in filters:
        if fil is dict:
            # Handle variable casing from past versions: only support snake evnetually
            column_name = fil.get("column_name", fil.get("columnName"))
            operator = fil.get("operator")
            comparison_value = fil.get("comparison_value", fil.get("comparisonValue"))
            compound_boolean = fil.get("compound_boolean", fil.get("compoundBoolean", compound_boolean))

            # Handle override operators
            if operator not in ALLOWED_OPERATORS:
                raise_exception(f"operator {operator} is not supported")
            if operator == "CONTAINS":
                operator = "LIKE"

            filter_string += f"{compound_boolean} {column_name} {operator} {comparison_value} \n"
        elif fil is str and fil != "":
            filter_string += f"{compound_boolean} {fil}"
```

kimballh (changes requested):
I ran the tests on plot and had to tweak a few things. I guess in some places, the macro gets a list of strings which screws things up. This passed all the tests though.

In the same pr that updates RasgoTransforms in the API, we will also need to remove the import statement from the `select_top_rows_cte.sql` template.

kimballh (approved):
Glad I ran the tests again, forgot a space, but I pushed that tiny change and I think it's good to ship now! Don't forget to bump the version.

PR #341, opened 2022-12-27
Make pivot great (and call it pivot_table)

cpdough:


cpdough:
working in testing on snowflake. haven't tested on BQ.

wish list:
- name the pivoted columns better (include the aggregation and values column name)
- support a UI type that combines the column and agg into a single argument

cpdough:
> working in testing on snowflake. haven't tested on BQ.
> 
> wish list:
> 
> name the pivoted columns better (include the aggregation and values column name)
support a UI type that combines the column and agg into a single argument
support count distinct aggregation


update... only to-dos left are test on BQ and support a UI type that combines the column and agg into a single argument

PR #355, opened 2023-01-22
add vlookup transform

cpdough:
we <3 excel

### DIFF
--- a/rasgotransforms/rasgotransforms/transforms/vlookup/vlookup.sql
+++ b/rasgotransforms/rasgotransforms/transforms/vlookup/vlookup.sql
@@ -1,0 +1,27 @@
+{%- macro table_from_fqtn(fqtn) -%}
+    {{ fqtn.split('.')[-1] }}
+{%- endmacro -%}
+
+{# Get all Columns in Source Table #}
+{%- set source_col_names = get_columns(source_table) -%}
+
+{# Get relevant Columns and Table Name in Lookup Table #}
+{%- if keep_columns is defined -%}
+    {%- set lookup_table_cols = keep_columns -%}
+{%- else -%}
+    {%- set lookup_table_cols = get_columns(lookup_table) -%}
+{%- endif -%}
+{%- set lookup_table_name = table_from_fqtn(lookup_table) -%}
+
+
+SELECT base.*, 
+{%- for column in lookup_table_cols %}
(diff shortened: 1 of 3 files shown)


rasgotransforms/rasgotransforms/transforms/vlookup/vlookup.yaml
```diff
@@ -0,0 +1,44 @@
+name: vlookup
+tags:
+  - table
+  - join
+  - reshape
+description: |
+  ## Inspired by Excel... a VLookup experience that works in SQL
+
+  ### Required Inputs
+  - Lookup Column: The column to look up in the Lookup Table. Make sure the column is named the same in both tables.
```
griffatrasgo, line 10: How soon do you anticipate users wanting to join on columns that don't have the same name? i.e. should we loosen the same name rule and allows for anchor_ & lookup_column args in the event they aren't named the same (which excel supports)? Or do we just push users to a regular join transform at that point?

rasgotransforms/rasgotransforms/transforms/vlookup/vlookup.sql
```diff
@@ -0,0 +1,27 @@
...
+{%- else -%}
+    {%- set lookup_table_cols = get_columns(lookup_table) -%}
+{%- endif -%}
+{%- set lookup_table_name = table_from_fqtn(lookup_table) -%}
+
+
+SELECT base.*, 
+{%- for column in lookup_table_cols %}
+    {%- if column in source_col_names -%}
+        lookup.{{ column }} as {{ lookup_table_name }}_{{ column }}{{ ', ' if not loop.last }}
```
griffatrasgo, line 20: nice touch 👍

griffatrasgo (approved):
Nice transform. Good kickoff to Rasgo's "Excel phase" :)
As a dumb user, I could see value in supporting different column names across the tables, but i defer to the author

rasgotransforms/rasgotransforms/transforms/vlookup/vlookup.yaml
cpdough: yeah i was wondering the same thing, but i tried to stay true to how vlookup does the lookup and it doesn't have a concept of "mapping" one column name to another (it just goes by position)

PR #362, opened 2023-02-08
make publish action manual

kimballh:
We have a problem when a transform updates requires the rasgotransform package to be updated in the api before it can be used, but merging the PR will automatically publish them. So, this updates the actions to be manually dispatched. It accepts an input for publishing to a specific environment ('staging', 's3', 'production', or 'all'). This change will also allow us to publish transforms to staging when we want to test there before publishing to prod.


griffatrasgo (approved):
Nice usability changes!

PR #367, opened 2023-02-12
Clean Up old transforms

cpdough:
these SQL functions are better suited for the new 'New Columns' transform rather than custom random rasgo transforms. let's kill them


griffatrasgo (approved):
-258 == +1!

